PR #6, opened 2023-05-18
feat: improve error handling and code structure

veasman:
This PR makes some structural changes and general code changes in order to enhance the error handling of the plugin.

**NOTE:**

Likely due to a mix of spaces and tabs, GitHub displays the code to be of entirely different indentations. This will most likely not be an issue when opening the code in your editor of choice.

Summary of changes made:

### init.lua

Redundant Code: Code for building binary was extracted out into a function to simplify code structure.

Error Handling: 'project_id' is never checked to be nil. If a nil project_id was passed the function would throw an error, but it would not halt or return the function.

Checking Binary Existence: Checking the existence of the binary is done twice using two different methods. The code was changed to use one method for consistency.

Exception Messages: Improvements were made where exceptions were thrown to increase detail for an easier time debugging down the time.

### main.go:

Error Reporting: 'log.Fatalf' already terminates program. Redundant os.Exit(1) was removed

Error Context: While 'errCheck' is a good simplification, it can create difficulty determining the error's source as this plugin scales since the context is lost.

Encapsulate Switch Statement: This was done mostly for visual clarity's sake.


harrisoncramer:
Thank you very much for writing up an MR! I'm in the process of actually moving this repository over to an HTTP-based approach (just merged in an MR that does most of the work for this). Some of your above contributions are definitely still valuable, if you want to resolve the conflicts then I'm happy to merge it in.

veasman:
All conflicts have been resolved, and mixed spacing should be fixed as well.

Please let me know if you'd like to see any further changes in this merge.

PR #35, opened 2023-08-15
Fixes #34 add path seperator for windows 

### LINKED ISSUE #34: Cant install on Windows
## Prerequsities

- [x] The "Troubleshooting" section of the README did not help

## Bug Description

When I PackerSync I only get the Message 'Could not install gitlab.nvim'.
at one point there is another error wich says : "C:\PathToPacker\gitlab.nvim/bin' is not recognized as an internal or external command,^M" 
notice the different slashes.
I am on a windows machine perhaps thats the issue? 

## Reproduction Steps

1. Add use 'harrisoncramer/gitlab.nvim' to config file
2. :PackerSync

Umzugskarton:


Umzugskarton:
I havent tested the changes completely yet

harrisoncramer:
This would be great, if you're willing to put up an MR fixing these for Windows.

I don't have a Windows machine so would defer to you on the functionality of the plugin in a Windows environment. 

There are basically two files where this would be an issue, according to my grepping the codebase for a `/` in the code: In the `lua/gitlab/init.lua` file, where there are a couple of places where I assume the path separator is a `/`, and in `lua/gitlab/utils/init.lua` where I do the same thing.

Umzugskarton:
Yes I will try to fix it! :)

harrisoncramer:
I've made a bunch of changes, turns out I don't need the `branch_exists` function at all anymore, so the few small changes you've made should be all you really need.

harrisoncramer:
I implemented this in another branch, please feel free to comment here if you're having issues.

PR #54, opened 2023-09-03
Feat: See Pipeline Job Logs

harrisoncramer:
This MR lets you use `<leader>l` (the default keybinding) to trigger a "linewise" action when in the popup buffer. 

For the pipeline popup buffer, this grabs the logtrace (if it exists) for the current job and opens it in a new buffer.

### DIFF
--- a/lua/gitlab/actions/pipeline.lua
+++ b/lua/gitlab/actions/pipeline.lua
@@ -5,7 +5,10 @@ local Popup    = require("nui.popup")
 local state    = require("gitlab.state")
 local job      = require("gitlab.job")
 local u        = require("gitlab.utils")
-local M        = {}
+local M        = {
+  pipeline_jobs = nil,
+  pipeline_popup = nil
+}
 
 -- The function will render the Pipeline state in a popup
 M.open         = function()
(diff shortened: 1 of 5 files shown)


lua/gitlab/actions/pipeline.lua
```diff
@@ -73,6 +78,59 @@ M.retrigger    = function()
...
+    end
+
+    M.pipeline_popup:unmount()
+    vim.cmd.enew()
+
+    bufnr = vim.api.nvim_get_current_buf()
+    vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, lines)
+
+    local job_file_path = string.format("/tmp/gitlab.nvim.job-%d", j.id)
+    vim.cmd("w! " .. job_file_path)
```
harrisoncramer, line 126: This use of a temporary file and use of `term cat` let us use the ANSII escape codes to color/style the output of the logfile easily without having to parse out individual ANSII escape codes and so forth.

However, the use of the `/tmp` folder will probably be another breaking point of compatibility with Windows and should get picked up in https://github.com/harrisoncramer/gitlab.nvim/pull/35

PR #62, opened 2023-09-14
Prefer `head_pipeline` field when getting pipeline info

johnybx:
Hi,
I noticed that sometimes I cannot get pipeline popup because `pipeline` in `state.INFO.pipeline` is `nil`. Based on [docs](https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr) the field `head_pipeline` should be used instead. I updated the code so that it prefer `head_pipeline` and fallback to `pipeline`.


harrisoncramer:
Thank you for doing this!

PR #65, opened 2023-10-12
Add option to choose random server port

johnybx:
Hi,
currently `gitlab.nvim` can be used only in one instance of neovim because there is only single port. Also when you open second neovim instance the server is not started silently (`http.Server` fails in go but the process still runs) but review is still opened and all API calls from second instance are sent to server which was started by first instance of neovim ( likely different gitlab project ).

I updated the start logic in lua and go in order to support random port which is chosen in go. This change is backward compatible so the `port` config option can still be used to specify port on which server should be started. Serve now have health check endpoint (`/ping`) and review is started only if this endpoint respond with HTTP status 200.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Here is the default setup function. All of these values are optional, and if you
 
 ```lua
 require("gitlab").setup({
-  port = 21036, -- The port of the Go server, which runs in the background, if omitted random port will be chosen automatically.
+  port = nil, -- The port of the Go server, which runs in the background, if omitted or `nil` the port will be chosen automatically
   log_path = vim.fn.stdpath("cache") .. "/gitlab.nvim.log", -- Log path for the Go server
   reviewer = "delta", -- The reviewer type ("delta" or "diffview")
   attachment_dir = nil, -- The local directory for files (see the "summary" section)
--- a/README.md
+++ b/README.md
@@ -13,7 +13,6 @@ And a lot more!
 
 https://github.com/harrisoncramer/gitlab.nvim/assets/32515581/50f44eaf-5f99-4cb3-93e9-ed66ace0f675
 
-
 ## Requirements
 
 - <a href="https://go.dev/">Go</a> >= v1.19
@@ -22,10 +21,10 @@ https://github.com/harrisoncramer/gitlab.nvim/assets/32515581/50f44eaf-5f99-4cb3
 
 1. Install Go
 2. Install reviewer: <a href="https://github.com/dandavison/delta">delta</a> or <a href="https://github.com/sindrets/diffview.nvim">diffview</a>
-2. Add configuration (see Installation section)
-3. Checkout your feature branch: `git checkout feature-branch`
-4. Open Neovim
-5. Run `:lua require("gitlab").review()` to open the reviewer pane
+3. Add configuration (see Installation section)
+4. Checkout your feature branch: `git checkout feature-branch`
+5. Open Neovim
+6. Run `:lua require("gitlab").review()` to open the reviewer pane
 
 ## Installation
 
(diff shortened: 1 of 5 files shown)


harrisoncramer:
Thank you for the contribution Johnny. 

Appreciate you noting that the server will reach out to the old server and cause issues.

However, your current solution is not respecting an initial port setting. I'd expect that if I set a value for `port` in the setup function, that port value would be used for the first server, and that _subsequent_ port values would be chosen at random. That is not happening, your changes are setting a random port no matter what value I provide in the settings.

It looks like this is not working because you are now providing `nil` as the initial value of the port in the settings module. The `merge` function that merges the initial settings and the new settings will not set the port value when that's the case. I think that's happening because the `pairs` function in Lua ignores `nil` values...

harrisoncramer:
The plugin wasn't really designed for multiple _different_ servers running on different projects at the same time, the `port` setting right now is just a single option. 

If this is something you want to support (and it seems like it could be useful generally) I think we should move the `port` option out of the `setup` function (since it's no longer globally applicable) and into the `.gitlab.nvim` file, which is supposed to be project specific. Alternatively, we should just nix the ability to set the specific port, and just surface it in the settings object after startup. For instance even with your current approach you can see the port of the current server like this:

```lua
:lua print(require("gitlab").state.settings.port)
```

johnybx:
Thanks for review. I didn't notice I broke previous behavior ( probably updated defaults after testing ). My intention was to keep both behaviors because some users might already be used to existing behavior. I didn't want to move the configuration to `.gitlab.nvim` because I personally don't see reason why I should care about the port on which the server runs on localhost :thinking: ( I don't think it makes sense to implement unless someone specifically asks for that functionality ).

>  I'd expect that if I set a value for port in the setup function, that port value would be used for the first server, and that subsequent port values would be chosen at random.

This is interesting idea but that would require checking if port is already taken because separate instances of neovim does not interact with each other :thinking: but also I am not sure what would be added value of running  first server on specific port.

Would you mind accepting this PR if I will fix the previous behavior? So users could set specific port for server or keep it `nil` and port would be chosen by random. Changes in `main.go` should also make sure that user will be notified that server was not started properly because port is already taken.

harrisoncramer:
Thank you very much for working on this. Just to be clear, is what you're suggesting:

1. Users can either set no port in their configuration for `port`, in which case, the server will always open a random port value.
2. A specific port. This would prevent a second Neovim instance from opening up a server because that port is already in use.

Is that right? And if so, that sounds fine to me. Let me know if that's right. and I can re-try your MR once you've gotten it working...

johnybx:
yes that is it :+1: . I will try to get to implementing it during the weekend. Thanks.

johnybx:
@harrisoncramer the issue with port should be fixed. Now when you specify port and try to open second instance of reviewer the error with the corresponding message is shown to user and reviewer is not opened. When you set port to `nil` or not specify at all a random port is chosen.

lua/gitlab/utils/init.lua
```diff
@@ -126,20 +126,10 @@ M.create_popup_state = function(title, width, height)
...
-  for key, value in pairs(defaults) do
-    if type(value) == "table" then
-      result[key] = M.merge(value, overrides[key] or {})
-    else
-      result[key] = overrides[key] or value
-    end
-  end
-
-  return result
+  return vim.tbl_deep_extend("force", defaults, overrides)
```
harrisoncramer, line 132: Wow nice, did not know about this function before.

README.md
```diff
@@ -85,7 +84,7 @@ Here is the default setup function. All of these values are optional, and if you
 
 ```lua
 require("gitlab").setup({
-  port = 21036, -- The port of the Go server, which runs in the background
+  port = 21036, -- The port of the Go server, which runs in the background, if omitted random port will be chosen automatically.
```
harrisoncramer, line 87: Let's write this as `nil` in the documentation since that's what's used by default:

```
 port = nil, -- The port of the Go server, which runs in the background, if omitted or `nil` the port will be chosen automatically
```

harrisoncramer:
This works great, left one small suggestion on the README, and once that's changed I'll squash and merge. Thank you for this contribution 🎉

README.md
johnybx: you mean `nil` in comment or instead of value 21036?

PR #68, opened 2023-10-28
Feat: Improve Logging for Go Server

harrisoncramer:
This MR improves the logging for the Go server by allowing users to log out the full request or response bodies in the setup function, aiding in debugging connectivity issues and other errors with the client-server data flow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Here is the default setup function. All of these values are optional, and if you
 require("gitlab").setup({
   port = nil, -- The port of the Go server, which runs in the background, if omitted or `nil` the port will be chosen automatically
   log_path = vim.fn.stdpath("cache") .. "/gitlab.nvim.log", -- Log path for the Go server
-  debug_type = "", -- "" (nothing), "request", "response", or "both" for the Go server logs
+  debug = { go_request = false, go_response = false }, -- Which values to log
   reviewer = "delta", -- The reviewer type ("delta" or "diffview")
   attachment_dir = nil, -- The local directory for files (see the "summary" section)
   popup = { -- The popup for comment creation, editing, and replying
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ Here is the default setup function. All of these values are optional, and if you
 require("gitlab").setup({
   port = nil, -- The port of the Go server, which runs in the background, if omitted or `nil` the port will be chosen automatically
   log_path = vim.fn.stdpath("cache") .. "/gitlab.nvim.log", -- Log path for the Go server
+  debug_type = "", -- "" (nothing), "request", "response", or "both" for the Go server logs
   reviewer = "delta", -- The reviewer type ("delta" or "diffview")
   attachment_dir = nil, -- The local directory for files (see the "summary" section)
   popup = { -- The popup for comment creation, editing, and replying
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -1,11 +1,11 @@
 package main
 
 import (
-	"bufio"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
+	"net/http/httputil"
 	"os"
 	"strconv"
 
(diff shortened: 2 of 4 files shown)


harrisoncramer:
@johnybx Would love to get your review of this. Think more optional, more verbose logging will be helpful for people and us trying to debug issues in the future.

cmd/client.go
```diff
@@ -17,23 +19,75 @@ type Client struct {
...
 	authToken      string
+	logPath        string
+	debug          bool
 	git            *gitlab.Client
 }
 
-type Logger struct {
-	Active bool
+var requestLogger retryablehttp.RequestLogHook = func(l retryablehttp.Logger, r *http.Request, i int) {
+	logPath := os.Args[len(os.Args)-1]
```
johnybx, line 28: I would maybe consider creating function which will accept `Client` and return  `retryablehttp.RequestLogHook` just so that we read args only in one place. 

It should also be possible to get context from `request` and `response.Request` ? But not sure if that is reliable :thinking:

README.md
```diff
@@ -86,6 +86,7 @@ Here is the default setup function. All of these values are optional, and if you
 require("gitlab").setup({
   port = nil, -- The port of the Go server, which runs in the background, if omitted or `nil` the port will be chosen automatically
   log_path = vim.fn.stdpath("cache") .. "/gitlab.nvim.log", -- Log path for the Go server
+  debug_type = "", -- "" (nothing), "request", "response", or "both" for the Go server logs
```
johnybx, line 89: I would maybe go for something like
```lua
debug = { go_request = false,  go_response = false }
```
so that it is  more extensible. The reason for that is that maybe later it would be also nice to have some debug / trace level of logging in lua.

johnybx:
Hi,
I had just 2 small comments otherwise it looks great :+1:. I tested it and yes this would be very useful for debugging multiline comments :grin:.

BTW just so that you know, my last two PRs were literally first lines of go code which I ever wrote so maybe I will not notice everything :slightly_smiling_face:

cmd/client.go
```diff
@@ -17,23 +19,75 @@ type Client struct {
...
+	}
+	defer file.Close()
+
+	clonedRequest := r.Clone(r.Context())
+	var data []byte
+	reader := bufio.NewReader(clonedRequest.Body)
+	_, err = reader.Read(data)
+	if err != nil {
+		panic(err)
+	}
```
johnybx, line 42: Looks like this does not actually read data. I tried this:
```suggestion
	var data []byte
	if clonedRequest.Body != nil {
		data, err = io.ReadAll(clonedRequest.Body)
		if err != nil {
			panic(err)
		}
	}
```
and it reads data. Or we could consider https://pkg.go.dev/net/http/httputil#DumpRequest.

cmd/client.go
```diff
@@ -17,23 +19,75 @@ type Client struct {
...
 	}
 	defer file.Close()
-	_, err = file.Write([]byte(logString))
+
+	var data []byte
+	reader := bufio.NewReader(response.Body)
+	_, err = reader.Read(data)
+	if err != nil {
+		panic(err)
+	}
```
johnybx, line 76: This is similar to previous comment.
```suggestion
	var data []byte
	if response.Body != nil {
		data, err = io.ReadAll(response.Body)
		if err != nil {
			panic(err)
		}
		response.Body = io.NopCloser(bytes.NewBuffer(data))
	}
```
Here we cannot leave response.Body empty ( already read ) otherwise handler will not be able to get content and will return error.

There is also alternative in httputil package https://pkg.go.dev/net/http/httputil#DumpResponse

johnybx (commented):
I was playing around with logging and wrote into comments what I found. If we don't want to pretty format json body then httputil package is probably easier to use.

README.md
harrisoncramer: Made that change.

PR #78, opened 2023-11-01
Discussion sign and diagnostics

johnybx:
:wave: Hi,
this PR adds support for discussion sign and diagnostics. Diagnostics and signs are currently split on purpose ( even though the sign can be set also for diagnostic ) to have better control when multiline support will be added ( in this PR ). I will be testing this for few days to see if it works.

TODO:
* :heavy_check_mark: support multiline discussions for sign -> the idea is to have some main sign on line where the discussion is created and then other sign(s) ( maybe something like <--> but vertical ) if discussion is for multiple lines.
* :heavy_check_mark: support multiline discussions for diagnostics.
* currently the discussions are refreshed when file is switched / loaded in reviewer -> do we need periodic updates ? it would be maybe convenient :thinking: 
    * :heavy_check_mark:  ~~if we will have period updates maybe it is not great idea to always remove everything and set it again -> it is not hard to update to incremental updates.~~ this does not seem to be needed
* :heavy_check_mark: ~~actions for diagnostics / sign lines ( this may even be new PR )~~  for this PR I added function to jump to discussion tree where all other discussion actions are available. 
*  :heavy_check_mark: Maybe we should show sign and diagnostic for notes which match checksum ? Otherwise the discussion will have likely incorrect position :thinking:

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,10 @@ https://github.com/harrisoncramer/gitlab.nvim/assets/32515581/50f44eaf-5f99-4cb3
 ## Quick Start
 
 1. Install Go
-3. Add configuration (see Installation section)
-4. Checkout your feature branch: `git checkout feature-branch`
-5. Open Neovim
-6. Run `:lua require("gitlab").review()` to open the reviewer pane
+2. Add configuration (see Installation section)
+3. Checkout your feature branch: `git checkout feature-branch`
+4. Open Neovim
+5. Run `:lua require("gitlab").review()` to open the reviewer pane
 
 ## Installation
 
@@ -66,7 +66,7 @@ use {
 
 ## Project Configuration
 
-This plugin requires an <a href="https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html#create-a-personal-access-token">auth token</a> to connect to Gitlab. The token can be set in the root directory of the project in a `.gitlab.nvim` environment file, or can be set via a shell environment variable called `GITLAB_TOKEN` instead. If both are present, the `.gitlab.nvim` file will take precedence. 
+This plugin requires an <a href="https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html#create-a-personal-access-token">auth token</a> to connect to Gitlab. The token can be set in the root directory of the project in a `.gitlab.nvim` environment file, or can be set via a shell environment variable called `GITLAB_TOKEN` instead. If both are present, the `.gitlab.nvim` file will take precedence.
 
 Optionally provide a GITLAB_URL environment variable (or gitlab_url value in the `.gitlab.nvim` file) to connect to a self-hosted Gitlab instance. This is optional, use ONLY for self-hosted instances.
 
@@ -112,6 +112,23 @@ require("gitlab").setup({
     resolved = '✓', -- Symbol to show next to resolved discussions
     unresolved = '✖', -- Symbol to show next to unresolved discussions
   },
+  discussion_sign = {
+    -- See :h sign_define for details about sign configuration.
+    enabled = true,
+    text = "💬",
+    linehl = nil,
+    texthl = nil,
+    culhl = nil,
+    numhl = nil,
+  },
+  discussion_diagnostics = {
+    -- If you want to customize diagnostics for discussions you can make special config
+    -- for namespace `gitlab_discussion`. See :h vim.diagnostic.config
+    enabled = true,
+    severity = vim.diagnostic.severity.INFO,
+    code = nil, -- see :h diagnostic-structure
+    display_opts = {}, -- see opts in vim.diagnostic.set
+  },
   pipeline = {
     created = "",
     pending = "",
@@ -145,7 +162,7 @@ Then open Neovim. To begin, try running the `summary` command or the `review` co
 
 ### Summary
 
-The `summary` action will open the MR title and description. 
+The `summary` action will open the MR title and description.
 
 ```lua
 require("gitlab").summary()
@@ -186,6 +203,14 @@ If you'd like to create a note in an MR (like a comment, but not linked to a spe
 
 ```lua
 require("gitlab").create_note()
+```
+
+### Discussions signs and diagnostics
+
+By default when reviewing files you will see signs and diagnostics ( if enabled in configuration ). When cursor is on diagnostic line you can view discussion thread by using `vim.diagnostic.show`. You can also jump to discussion tree where you can reply, edit or delete discussion.
+
+```lua
+require("gitlab").jump_to_discussion_tree_from_diagnostic()
 ```
 
 ### Uploading Files
@@ -249,6 +274,7 @@ vim.keymap.set("n", "<leader>glR", gitlab.revoke)
 vim.keymap.set("n", "<leader>glc", gitlab.create_comment)
 vim.keymap.set("v", "<leader>glc", gitlab.create_multiline_comment)
 vim.keymap.set("v", "<leader>glC", gitlab.create_comment_suggestion)
+vim.keymap.set("n", "<leader>glj", gitlab.jump_to_discussion_tree_from_diagnostic)
 vim.keymap.set("n", "<leader>gln", gitlab.create_note)
 vim.keymap.set("n", "<leader>gld", gitlab.toggle_discussions)
 vim.keymap.set("n", "<leader>glaa", gitlab.add_assignee)
--- a/README.md
+++ b/README.md
@@ -228,12 +228,10 @@ require("gitlab").move_to_discussion_tree_from_diagnostic()
 
 The `discussion_sign` configuration controls the display of signs for discussions. The `enabled` option turns on/off the signs. `text` sets the sign text. `linehl`, `texthl`, `culhl`, and `numhl` customize the line highlighting, text highlighting, column highlighting, and number highlighting respectively. Keep in mind that these can be overridden by other configuration (for example diffview.nvim highlights). `priority` controls the sign priority order (when multiple signs are placed on the same line, the sign with highest priority is used). The `helper_signs` table configures additional signs for multiline discussions in order to show the whole context. `enabled` turns on/off the helper signs. `start`, `mid`, and `end` set the helper sign text.
 
-The `discussion_diagnostic` configuration customizes the diagnostic display for discussions. The `enabled` option turns on/off the diagnostics. `severity` sets the diagnostic severity level and should be set to one of `vim.diagnostic.severity.ERROR`, `vim.diagnostic.severity.WARN`, or `vim.diagnostic.severity.INFO`, `vim.diagnostic.severity.HINT`. `code` specifies a diagnostic code. `display_opts` configures the diagnostic display options where you can configure values like:
+The `discussion_diagnostic` configuration customizes the diagnostic display for discussions. The `enabled` option turns on/off the diagnostics. `severity` sets the diagnostic severity level and should be set to one of `vim.diagnostic.severity.ERROR`, `vim.diagnostic.severity.WARN`, or `vim.diagnostic.severity.INFO`, `vim.diagnostic.severity.HINT`. `code` specifies a diagnostic code. `display_opts` configures the diagnostic display options where you can configure values like (this is dirrectly used as opts in vim.diagnostic.set):
 
 - `virtual_text` - Show virtual text for diagnostics.
 - `underline` - Underline text for diagnostics.
-- `severity_sort` - Sort diagnostics by severity.
-- `float` - Show diagnostics in a floating window.
 
 Diagnostics for discussions use the `gitlab_discussion` namespace. See `:h vim.diagnostic.config` and `:h diagnostic-structure` for more details.
 
--- a/README.md
+++ b/README.md
@@ -112,6 +112,10 @@ require("gitlab").setup({
     resolved = '✓', -- Symbol to show next to resolved discussions
     unresolved = '✖', -- Symbol to show next to unresolved discussions
   },
+  discussion_sign_and_diagnostic = {
+    skip_resolved_discussion = false,
+    skip_old_revision_discussion = true,
+  },
   discussion_sign = {
     -- See :h sign_define for details about sign configuration.
     enabled = true,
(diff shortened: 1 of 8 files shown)


lua/gitlab/reviewer/init.lua
```diff
@@ -37,6 +37,31 @@ M.init = function()
 
   M.get_lines = reviewer.get_lines
   -- Returns the content of the file in the current location in the reviewer window
+
+  M.get_current_file = reviewer.get_current_file
+  -- Get currently loaded file
+
+  M.place_sign = reviewer.place_sign
+  -- Places a sign on the line for currently reviewed file.
+  -- Prameters:
```
harrisoncramer, line 46: Typos (grep for prameters)

README.md
```diff
@@ -252,6 +290,7 @@ vim.keymap.set("n", "<leader>glR", gitlab.revoke)
 vim.keymap.set("n", "<leader>glc", gitlab.create_comment)
 vim.keymap.set("v", "<leader>glc", gitlab.create_multiline_comment)
 vim.keymap.set("v", "<leader>glC", gitlab.create_comment_suggestion)
+vim.keymap.set("n", "<leader>glj", gitlab.jump_to_discussion_tree_from_diagnostic)
```
harrisoncramer, line 293: I'd recommend `glm` (`move_to_discussion_tree_from_diagnostic`) as the default here to match the default `m` used to move to the diagnostic from the discussion tree. Easier to remember IMO.

lua/gitlab/actions/discussions.lua
```diff
@@ -14,32 +14,401 @@ local miscellaneous = require("gitlab.actions.miscellaneous")
...
+  reviewer.set_diagnostics(
+    diagnostics_namespace,
+    old_diagnostics,
+    "old",
+    state.settings.discussion_diagnostics.display_opts
+  )
+end
+
+---Refresh discussion data, discussion signs and diagnostics
+M.refresh_discussion_data = function()
```
harrisoncramer, line 356: Can we please add an invocation of this function in the `lua/gitlab/actions/comment.lua` file, after both notes and comments are created? This will make the comment appear in the buffer right away.

lua/gitlab/state.lua
```diff
@@ -33,6 +33,32 @@ M.settings = {
...
+    helper_signs = {
+      -- For multiline comments the helper signs are used to indicate the whole context
+      -- Priority of helper signs is lower than the main sign (-1).
+      enabled = true,
+      start = "↑",
+      mid = "|",
+      ["end"] = "↓",
+    },
+  },
+  discussion_diagnostics = {
```
harrisoncramer, line 54: Sort of the same note as before, diagnostics are a pretty niche part of the Neovim experience, if we can add a short explanation into the README that would be great. Otherwise I don't know that it will be very user-friendly, and doubt that folks will actually use it.

harrisoncramer (changes requested):
This is incredibly awesome, left some comments!

README.md
```diff
@@ -109,6 +109,32 @@ require("gitlab").setup({
...
+    culhl = nil,
+    numhl = nil,
+    priority = 20,
+    helper_signs = {
+      -- For multiline comments the helper signs are used to indicate the whole context
+      -- Priority of helper signs is lower than the main sign (-1).
+      enabled = true,
+      start = "↑",
+      mid = "|",
+      ["end"] = "↓",
```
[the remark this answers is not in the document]
johnybx: This can actually happen when you create visual selection and press `o` to jump to other end of the selection you can actually create comment which will have sign at start and you would see end arrow like this 
image

README.md
```diff
@@ -109,6 +109,32 @@ require("gitlab").setup({
     resolved = '✓', -- Symbol to show next to resolved discussions
     unresolved = '✖', -- Symbol to show next to unresolved discussions
   },
+  discussion_sign = {
+    -- See :h sign_define for details about sign configuration.
+    enabled = true,
+    text = "💬",
+    linehl = nil,
```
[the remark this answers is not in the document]
johnybx: I see, I will try to add also callback to remove highlights and signs when reviewer is closed or left ( will see what options there are available ).

Regarding the `linehl` you are right that highlight is applied only on unchanged lines ( if there is comment ) - I need to check if it is possible to define some highligh priority or something like that. The idea here was to make all settings available to user but I didn't try all of them. If nothing more it might be at least useful to see context of comment outside of changed lines :thinking:

lua/gitlab/actions/discussions.lua
```diff
@@ -14,32 +14,401 @@ local miscellaneous = require("gitlab.actions.miscellaneous")
 
 local edit_popup = Popup(u.create_popup_state("Edit Comment", "80%", "80%"))
 local reply_popup = Popup(u.create_popup_state("Reply", "80%", "80%"))
+local discussion_sign_name = "gitlab_discussion"
+local discussion_helper_sign_start = "gitlab_discussion_helper_start"
+local discussion_helper_sign_mid = "gitlab_discussion_helper_mid"
+local discussion_helper_sign_end = "gitlab_discussion_helper_end"
+local diagnostics_namespace = vim.api.nvim_create_namespace(discussion_sign_name)
+
+vim.fn.sign_define(discussion_sign_name, {
```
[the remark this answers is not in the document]
johnybx: good point, updated the code :+1:

README.md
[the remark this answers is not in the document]
johnybx: The problem with signs and diagnostics outside of reviewer is fixed.

lua/gitlab/reviewer/init.lua
johnybx: fixed.

README.md
johnybx: updated.

README.md
```diff
@@ -191,6 +217,18 @@ If you'd like to create a note in an MR (like a comment, but not linked to a spe
 require("gitlab").create_note()
 ```
 
+### Discussions signs and diagnostics
```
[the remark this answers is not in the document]
johnybx: I updated the docs, please check if it is sufficient.

lua/gitlab/state.lua
johnybx: I updated the docs, please check if it is sufficient.

README.md
```diff
@@ -191,6 +221,34 @@ If you'd like to create a note in an MR (like a comment, but not linked to a spe
...
+
+The `discussion_diagnostic` configuration customizes the diagnostic display for discussions. The `enabled` option turns on/off the diagnostics. `severity` sets the diagnostic severity level and should be set to one of `vim.diagnostic.severity.ERROR`, `vim.diagnostic.severity.WARN`, or `vim.diagnostic.severity.INFO`, `vim.diagnostic.severity.HINT`. `code` specifies a diagnostic code. `display_opts` configures the diagnostic display options where you can configure values like:
+
+- `virtual_text` - Show virtual text for diagnostics.
```
harrisoncramer, line 236: Perhaps I'm doing something wrong, but these options don't seem to be applying for me... The only one that's working is setting `virtual_text` to a boolean value.

harrisoncramer (commented):
Left a few more recommendations. This is looking very close. Nice work!

lua/gitlab/actions/discussions.lua
```diff
@@ -51,24 +458,91 @@ M.toggle = function()
...
-      { unlinked_section.bufnr, data.unlinked_discussions, "No Notes (Unlinked Discussions) for this MR" },
+  if #diagnostics == 0 then
+    vim.notify("No diagnostics for this line", vim.log.levels.WARN)
+    return
+  elseif #diagnostics > 1 then
+    local lines = {}
+    for _, diagnostic in pairs(diagnostics) do
+      table.insert(lines, Menu.item(diagnostic.user_data.header, { diagnostic = diagnostic }))
+    end
+    local menu = Menu({
```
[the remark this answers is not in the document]
harrisoncramer: (I've got to move the delete comment confirmation dialogue modal over too!)

README.md
johnybx: I tried both `{virtual_text = false, underline = false}` and these works for me -> these were main reason for having `display_opts` to be able to have different settings than regular diagnostics.

I think it makes sense to remove `float` because that is basically configuration for floating window ( table ) and probably too confusing here (`:h vim.diagnostic.open_float`). Same goes for `severity_sort` because if works but only within our defined namespace so I understand the confusion here. Advanced users will be still able to put there any settings from `:h vim.diagnostic.config`

lua/gitlab/actions/discussions.lua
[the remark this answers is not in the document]
johnybx: sure, looks much easier :+1:

README.md
johnybx: but the thing with float is interesting because this respect settings
```
:lua vim.diagnostic.open_float({header = "Hi",       source = false,        border = "rounded",})
```
but my shortcut does not :thinking:

harrisoncramer: Yeah that's what I was noticing, I was hoping to be able to set specific float options for Gitlab (e.g. title, formatting the text, etc) but they weren't working for me.

johnybx: It seems that I cannot make the float work. One thing is that I have key binding with specific settings which should override anything:
```
    "n",
    "<leader>cd",
    "<cmd>lua vim.diagnostic.open_float(0, {scope='line', border = 'rounded'})<CR>",
    { silent = true }
```

but even when I use `open_float` without config it does not respect settings. I can try to dig into that in following days :thinking:

harrisoncramer: That's fine, let's not worry about it as part of this MR. Perhaps worth asking about in other forums.

harrisoncramer:
I'm good to merge this @johnybx, is that okay with you? Let me know and will do it.

harrisoncramer:
Oh actually, an issue I've noticed -- for some reason when you try and use any other command from `diffview` such as `DiffviewFileHistory` _outside_ of this plugin, it's now triggering your code to `refresh_discussion_data` which is causing an error!

johnybx:
Ah I see, I think I can easily fix it by setting up autocommands at open and then doing cleanup on close. Other option would be to actually check which instance of diffview triggered the autocommand and only conditionally react - that would work even if you opened multiple instances of diffview. I will try to implement it this week :+1:

cmd/git.go
```diff
@@ -19,7 +19,19 @@ func ExtractGitInfo() (string, string, string, string, error) {
...
+	           https://gitlab.com/namespace/dummy-test-repo.git
+	           git@gitlab.com:namespace/dummy-test-repo.git
+	           ssh://git@gitlab.com/namespace/dummy-test-repo.git
+
+	       namespace: namespace/subnamespace, projectName: dummy-test-repo:
+	           ssh://git@gitlab.com/namespace/subnamespace/dummy-test-repo
+	           https://git@gitlab.com/namespace/subnamespace/dummy-test-repo.git
+	           git@git@gitlab.com:namespace/subnamespace/dummy-test-repo.git
+	*/
+	re := regexp.MustCompile(`(?:https?:\/\/|ssh:\/\/|git@)(?:[^\/:]+)[\/:](.*)\/([^\/]+?)(?:\.git)?$`)
```
johnybx, line 34: I found out that on many repositories I often use very different format of url ( I don't set it manually ) so I updated regexp to match urls in comment. I tested this manually and also [here](https://regex101.com/r/UU8IBo/1). The main problem was the assumption that namespace is just one word ( `([^\/]+` ) but namespace is everything in URI till the project name.

johnybx (commented):
I updated the auto commands to trigger actions only if it is related to diffview opened by gitlab.nvim and also tried having gitlab review and normal diffview and didn't noticed any errors. 
I'll be for sure doing multiple reviews next week so I will be testing more for sure :+1:

cmd/git.go
harrisoncramer: Yeah, good call, this is an awesome catch. I just merged in another contribution (https://github.com/harrisoncramer/gitlab.nvim/pull/87) that fixes this (as well as adding some other checks + tests) for the regex.

If you resolve the merge conflicts on your branch (with the latest merge) I'll merge this in. Thank you for your patience and for dealing with multiple conflicts!

harrisoncramer:
If you make my suggested changes and resolve the merge conflict then I'll merge this! Thank you for dealing with double conflicts here, appreciate this truly awesome contribution!

johnybx:
I resolved conflicts but I don't see your suggested changes :thinking:. I see only resolved comments - is it within resolved comments ?

harrisoncramer:
> I resolved conflicts but I don't see your suggested changes 🤔. I see only resolved comments - is it within resolved comments ?

I'm sorry, the review was never submitted, you should see them now

cmd/git.go
johnybx: I resolved the conflict here but my regexp is more extended - the previous version was for example missing `ssh://` and also the `.gitf

lua/gitlab/reviewer/diffview.lua
```diff
@@ -149,4 +161,73 @@ M.get_lines = function(start_line, end_line)
   return vim.api.nvim_buf_get_lines(0, start_line - 1, end_line, false)
 end
 
+---Get currently shown file
+M.get_current_file = function()
+  return diffview_lib.get_current_view().panel.cur_file.path
```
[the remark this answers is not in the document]
johnybx: added check :+1:

lua/gitlab/reviewer/diffview.lua
```diff
@@ -149,4 +161,73 @@ M.get_lines = function(start_line, end_line)
...
+  vim.fn.sign_placelist(signs)
+end
+
+---Set diagnostics in currently reviewed file.
+---@param namespace integer namespace for diagnostics
+---@param diagnostics table see :h vim.diagnostic.set
+---@param type string "new" if diagnostic should be in file after changes else "old"
+---@param opts table? see :h vim.diagnostic.set
+M.set_diagnostics = function(namespace, diagnostics, type, opts)
+  local view = diffview_lib.get_current_view()
```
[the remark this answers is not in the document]
johnybx: fixed.

cmd/git.go
johnybx: I also added tests.

PR #81, opened 2023-11-07
Breaking: Deprecate Delta Reviewer

harrisoncramer:
This MR removes the Delta reviewer.

It's my strong hunch that nobody is actually using this, and given that we are now adding features specific to Diffview we should deprecate this reviewer so that folks can have the full experience.

I've added guards into the application to alert users using Delta to switch over to Diffview.


harrisoncramer:
@johnybx Would you please review this? Should be a pretty straightforward deprecation of all things Delta.

lua/gitlab/actions/assignees_and_reviewers.lua
```diff
@@ -38,7 +38,7 @@ M.add_popup = function(type)
     table.insert(current_ids, choice.id)
     local body = { ids = current_ids }
     job.run_job("/mr/" .. type, "PUT", body, function(data)
-      vim.notify(data.message, vim.log.levels.INFO)
+      u.notify(data.message, vim.log.levels.INFO)
```
harrisoncramer, line 41: Adding this prefixes all messages with `gitlab.nvim:` which is great for usability IMO

lua/gitlab/init.lua
```diff
@@ -22,10 +22,11 @@ return {
     end
     server.build() -- Builds the Go binary if it doesn't exist
     state.setPluginConfiguration() -- Sets configuration from `.gitlab.nvim` file
-    state.merge_settings(args) -- Sets keymaps and other settings from setup function
+
+    if not state.merge_settings(args) then -- Sets keymaps and other settings from setup function
```
harrisoncramer, line 26: If the merge_settings function returns false it means there was a fatal error and we do not try to keep running the setup function

johnybx: shouldn't there be a notification for user ?

lua/gitlab/reviewer/init.lua
```diff
@@ -1,42 +0,0 @@
--- This Module will pick the reviewer set in the user's
```
johnybx, line 1: I personally would maybe leave this here because it nicely abstracts reviewer, we have control over functions which we use elsewhere in code ( assuming we use only things exposed here ) and it will allow for future extensibility if someone else want to support other reviewers ( or build some kind of plugin ). 

What do you think ? ( if you feel like it is bad idea feel free to ignore this :+1: )

johnybx (changes requested):
Minor comment but LGTM :+1:

lua/gitlab/init.lua
johnybx: my bad I just noticed that message is in function.

lua/gitlab/reviewer/init.lua
harrisoncramer: Sure, will put it back. No problem, thanks!

PR #83, opened 2023-11-08
Fix highlight for usernames

mrparalon:
1. Fixed regex for username. Now it supports`_-.` and Unicode symbols like German letters.
изображение
Here is example of usernames, which is valid, but didn't work before
```
@pötzsch 1 day ago 
@Aleksei.Sherchenkov 1 day ago 
@Aleksei-Sherchenkov 1 day ago 
@Aleksei_Sherchenkov-Second 1 day ago 
```

2. Fixed getting colors from HL groups. On my Mac in terminal, terminal using tmux and neovide it returned empty string. I took code from documentation of `vim.fn.hlID()`

### DIFF
--- a/after/syntax/gitlab.vim
+++ b/after/syntax/gitlab.vim
@@ -2,7 +2,7 @@ if filereadable($VIMRUNTIME . '/syntax/markdown.vim')
   source $VIMRUNTIME/syntax/markdown.vim
 endif
 
-syntax match Username "@\w\+"
+syntax match Username "@\I\+[\_\-\.]*\I\+"
 syntax match Date "\v\d+\s+\w+\s+ago"
 syntax match ChevronDown ""
 syntax match ChevronRight ""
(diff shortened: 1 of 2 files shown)


after/syntax/gitlab.vim
```diff
@@ -2,7 +2,7 @@ if filereadable($VIMRUNTIME . '/syntax/markdown.vim')
   source $VIMRUNTIME/syntax/markdown.vim
 endif
 
-syntax match Username "@\w\+"
+syntax match Username "@\I\+[\_\-\.]*\I\+"
```
harrisoncramer, line 5: What does the `\I` refer to here?

For me this isn't capturing the final digit in my username: 

Screenshot 2023-11-08 at 12 52 27 PM

harrisoncramer (changes requested):
Thank you for addressing this, one small comment, think we might have to stick with `\w` in your pattern of `\I` but not sure. I'm not a regex guru by any means.

after/syntax/gitlab.vim
mrparalon: FIxed

PR #84, opened 2023-11-09
Feat: Remove Requirement for Dotfile

harrisoncramer:
This MR makes it so that the `.gitlab.nvim` file is optional. If present, the variables provided in this file will override those from the user's environment. This should be a non-breaking change.

The API key and Github URL may now be provided as environment variables. The project ID is derived from an API call instead of provided as a variable. 

The configuration of the plugin with these variables is now deferred until the first command is actually run, because we need to make an additional API call to fetch the Project ID.


cmd/main.go
```diff
@@ -110,3 +115,21 @@ func PingHandler(w http.ResponseWriter, r *http.Request) {
...
+/* Gets the project name */
+func getProjectName() (res string, e error) {
+	hasRemote := exec.Command("git", "remote", "get-url", "origin")
+	_, err := hasRemote.Output()
+
+	if err != nil {
+		return "", fmt.Errorf("Could not get origin remote")
+	}
+
+	cmd := exec.Command("bash", "-c", "git remote get-url origin | xargs basename -s .git")
```
johnybx, line 128: gitlab API also allow you to use project path ( namespace and project name ) in API call for getting project information. Would be that more reliable ?

For example I use these aliases for `glab` cli tool to get project path , project info and id 
```
project_id	!glab project_info | jq ".id"
project_info	!glab api $(glab project_path | python -c "import sys,urllib.parse; input = sys.stdin.read(); input = input.strip(\"\n\"); print(f\"/projects/{urllib.parse.quote_plus(input)}\")")
project_path	!git ls-remote --get-url | python -c "import sys,re; input = sys.stdin.read(); m = re.match(\".*://git@[^/]*/(.*)|.*git@[^:]*:(.*)|https://[^/]*/(.*)\", input); print(m and (m.group(1) or m.group(2) or m.group(3)))"
```

It is little bit more parsing though. Or would you mind if we had support for used defined lua function which would return configuration ? That way users could define their own way how to get credentials ( I personally have alias `glab token` which returns token from glab settings based on gitlab instance for project.

harrisoncramer: So, the goal of this MR is to introduce one feature, as per the initial user's request in that ticket, which is to make the `.gitlab.nvim` file optional by providing shell variables. 

If we wanted to support a _third_ configuration option (the ability to call any arbitrary Lua code to evaluate environment variables) then that'd have to be done as a separate MR.

harrisoncramer: I've changed this to query for all projects which you are a member of, and then filter down to find the project whose SSH remote or HTTPS remote matches the origin remote of the current repository.

PR #87, opened 2023-11-12
Feat: support nested folders in namespace

l-lin:
Some of my projects are in nested folders, e.g. https://custom-gitlab.com/folder-1/folder-2/project-name.git.

The regex used seems to only support one level of folder.

This pull request changes the golang server to support multiple nested folders.
I also added some unit tests to ensure the regex is working as expected.
Moreover, I added some CI to build and test the golang project.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -1,0 +1,25 @@
+name: Go
+
+on:
+  push:
+    branches: [ "main" ]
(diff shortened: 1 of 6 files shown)


harrisoncramer:
Thank you, this is a great fix, and a great addition to the CI as well. Appreciate the contribution!!

PR #95, opened 2023-11-14
add option to resolve whole discussion thread

johnybx:
Hi :wave:,
I found out that currently there is option only to resolve single comment but in gitlab UI you can see that you still have option to resolve whole discussion thread. I added this functionality based on [gitlab API docs](https://docs.gitlab.com/ee/api/discussions.html#resolve-a-merge-request-thread). I wasn't sure if it is fine to rename `toggle_resolved` to `toggle_comment_resolved` but adding just new option `toggle_discussion_resolved` would be very confusing for user ( actually I am not really sure what is added value of resolving just single comment so maybe we could keep only one option :thinking: ).

BTW what do you use for formatting markdown? I use `prettier` and it auto-formatted a lot of the `README.md` so just wanted to check with you if it is OK or if I should disable auto-format and revert it :thinking:


harrisoncramer:
Thanks for adding this! I'll review it later tonight. Really appreciate all your contributions to the plugin.

Let's not worry about formatting the README, that feels like overkill.

cmd/resolve_discussion.go
```diff
@@ -0,0 +1,63 @@
...
+	)
+
+	if err != nil {
+		c.handleError(w, err, "Could not update resolve status of discussion", res.StatusCode)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	response := SuccessResponse{
+		Message: "Resolve status of discussion updated",
```
harrisoncramer, line 55: Can we be specific with this message? We have the boolean so we could easily write "Discussion resolved" or "Discussion unresolved" here.

lua/gitlab/init.lua
```diff
@@ -47,7 +47,8 @@ return {
   toggle_discussions = async.sequence({ info }, discussions.toggle),
   edit_comment = async.sequence({ info }, discussions.edit_comment),
   delete_comment = async.sequence({ info }, discussions.delete_comment),
-  toggle_resolved = async.sequence({ info }, discussions.toggle_resolved),
+  toggle_comment_resolved = async.sequence({ info }, discussions.toggle_comment_resolved),
```
harrisoncramer, line 50: Agreed with your comment! Can you please remove this endpoint, use the "p" keybinding to resolve the discussion instead of "P" and then remove the `toggle_comment_resolved` keybinding and logic

harrisoncramer: You can also simplify the PATCH `/comment` endpoint a little bit, by removing some of the other stuff:

```lua
	/* The PATCH can either be to the resolved status of
	the discussion or or the text of the comment */
	msg := "edit comment"
	if editCommentRequest.Comment == "" {
		options.Resolved = &editCommentRequest.Resolved
		msg = "update discussion status"
	} else {
		options.Body = gitlab.String(editCommentRequest.Comment)
	}
```

harrisoncramer (commented):
Hey, looks great, left a couple of small comments

cmd/resolve_discussion.go
johnybx: sure :+1:

lua/gitlab/init.lua
johnybx: I removed `toggle_comment_resolved` and updated the `PATCH` comment method. Should I keep the  option named `toggle_discussion_resolved` or you prefer `toggle_resolved` ?

harrisoncramer: We want peoples' keybindings (if they picked something other than "p") to automatically work, so let's keep the name as `toggle_resolved` instead. You can test it by adding a `settings.discussion_tree.toggle_resolved` keybinding.

johnybx: Updated :ok_hand: .

PR #100, opened 2023-11-16
Feat: Adds More Information to Summary Panel

harrisoncramer:
Adds new info panel to the summary view to give users more information about the current merge request:

Screenshot 2023-11-16 at 9 07 07 AM

This panel can be configured to be visible either underneath (default) the description, or to the left. Users can also turn it on or off, and configure which fields are displayed in it.


lua/gitlab/utils/init.lua
```diff
@@ -47,6 +47,33 @@ M.press_enter = function()
   vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("<CR>", false, true, true), "n", false)
 end
 
+M.offset_to_seconds = function(offset)
```
harrisoncramer, line 50: Hopefully this works in other timezones, it works perfectly for EST.

Basically we get the time from the MR, get the user's current timezone offset, convert that to seconds, and then add/subtract those seconds from the time of the MR (which is UTC). We then reconvert that numeric value back into a readable string. 

Parsing dates in Lua is difficult without any external dependencies!

PR #104, opened 2023-11-17
Fix: Go Action Runs on MR Only

harrisoncramer:
Cleans up Go action to only run on merge request into main


harrisoncramer (commented):
Approval

PR #106, opened 2023-11-18
Bugfix: Updates Diff Hashes

harrisoncramer:
Rather than using branch names we are using the hashes provided directly in the Gitlab API response, to compare the point at which the branch diverged from the target to the head commit. We are additionally flashing a warning if the MR contains a merge conflict.

This is a #patch change


lua/gitlab/reviewer/diffview.lua
```diff
@@ -10,8 +10,19 @@ local M = {
...
 
 M.open = function()
-  vim.api.nvim_command(string.format("DiffviewOpen %s", state.INFO.target_branch))
+  local diff_refs = state.INFO.diff_refs
+  if diff_refs == nil then
+    u.notify("Gitlab did not provide diff refs required to review this MR", vim.log.levels.ERROR)
+    return
+  end
+
+  vim.api.nvim_command(string.format("DiffviewOpen %s..%s", diff_refs.start_sha, diff_refs.head_sha))
```
harrisoncramer, line 19: Now rather than opening up with the two branch names we open up with the specific commits provided from Gitlab, after making sure they are available locally by pulling down the latest via `git fetch origin`

johnybx: @harrisoncramer I noticed that by doing this we loose a lot of goodies of neovim like LSP :thinking:.

Maybe it would be better to just do
```lua
vim.api.nvim_command(string.format("DiffviewOpen %s", diff_refs.base_sha))
```
and notify user if `diff_refs.head_sha` is not same as `git rev-parse HEAD` ? What do you think ?

PR #121, opened 2023-11-21
Feat: Add reviewers to summary metadata

jakubbortlik:
This commit makes it possible to show reviewers in the metadata panel of the Summary. It does not change the default configuration, though.


jakubbortlik:
I forgot to mention that I like your plugin a lot! Thanks!

harrisoncramer:
Thanks for adding this, and for the kind words! 

Can we add it to the default configuration, in fact? Better to have it there so folks can easily comment it out. Once you do that I'll approve and merge!

jakubbortlik:
I believe I've done all that is needed for "reviewers" to be in the summary by default.

PR #124, opened 2023-11-27
Feat: Support ISO 8601 timestamp with offset

mars90226:
According to Wikipedia, ISO 8601 format with timezone offset has the format like: `2021-01-01T00:00:00.000-05:00`.

This MR support this timestamp format for those GitLab server returning timestamp with offset.

Ref: https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC


harrisoncramer:
Thank you, this looks great!

If you format the branch with https://github.com/JohnnyMorganz/StyLua then I'll happily merge.

mars90226:
Done, thanks for the reminder.

PR #129, opened 2023-12-02
Make popups configurable

jakubbortlik:
Hi,

I've prepared a PR where I've made popup windows configurable - the user can now configure the border, width, height and opacity globally for all popups (`edit`, `reply`, `comment`, `note`, `pipeline`) and override any of the parameters for individual popups.

The main reason for this addition is that frequently when working with comments I need to see the context of the code I'm commenting on. So frequently I found myself just saving the contents of the opaque comment window to a register, closing the comment, inspecting the context, and creating the comment once again, and pasting the text back in.

Normally I don't need transparent popups, but in this case I find it quite useful, so ideally, I'd like to make the popups non-transparent most of the time and only toggle transparency when I need it. But I don't know how this could be achieved (i.e., how to update the settings of the currently visible popup).

I'm aware that there is #102, and there will be conflicts with my PR, so I'll wait till #102 is merged, ad then I'll fix the problems. Until then, @harrisoncramer you can tell me, whether you like this PR in the first place :) Mostly it does not affect the current default behaviour of `gitlab.nvim`, the only exception is that I've unified the default size of popups at `width="40%"` and `height="60%"`, the original values of `comment` and `note` popups, which changes how the `edit` and `reply` popups look (they were "80%" and "80%"). In my opinion, having just one default value makes user configuration easier.


harrisoncramer (commented):
Thank you for the MR! We can definitely make the popups configurable but I'd recommend taking a different approach that will make this easier to maintain, I've outlined the changes in the comments.

harrisoncramer:
@jakubbortlik We just merged #102 so whenever you fix the conflicts I'll be good to merge this in, thank you for your understanding.

jakubbortlik:
@harrisoncramer I've just merged in `main` and fixed the conflicts (and fixed some issues in the documentation along the way :)).

harrisoncramer:
What you did is fine! Thank you for the changes, just merged them in 🔥

PR #145, opened 2023-12-16
Feat: Skip Tags + Releases

harrisoncramer:
This will let us skip releases with the NONE keyword (no tags, no release) for small things like README tweaks or pipeline changes.

NONE


harrisoncramer:
This will let us skip releases with the NONE keyword (no tags, no release) for small things like README tweaks or pipeline changes. The NONE keyword will do this.

PR #146, opened 2023-12-16
Bugfix: Comments on Unchanged, Expanded Lines

harrisoncramer:


lua/gitlab/actions/discussions/signs.lua
```diff
@@ -125,7 +252,7 @@ M.refresh_signs = function(discussions)
         discussion_line = first_note.position.new_line
         start_line = start_new_line
         end_line = end_new_line
-      elseif first_note.position.line_range.start.type == "old" then
+      elseif start_type == "old" or start_type == "expanded" then
```
harrisoncramer, line 255: This is basically where we want to check the change. We can just use the old file if it's on an expanded line.

lua/gitlab/actions/discussions/signs.lua
```diff
@@ -136,7 +263,10 @@ M.refresh_signs = function(discussions)
         discussion_line = first_note.position.old_line
         start_line = start_old_line
         end_line = end_old_line
+      else
+        return {}, {}, string.format("Unsupported line range type found for discussion %d", discussion.id)
```
harrisoncramer, line 267: I'm adding this as a safeguard. Probably should not return but having an explicit error pattern is nice.

Issue #41, opened 2023-08-18
Feat: Create Merge Request


h3ndry:
"I've read the `README.md` document, but it only explains how to work with an already existing merge request. How can I create a new merge request if one doesn't exist?"

harrisoncramer:
This isn't currently supported, but I could add it if there's a lot of interest. It's not high on my list of priorities right now, though, since the [Gitlab CLI](https://gitlab.com/gitlab-org/cli) already supports this really well. For instance:

```
USAGE
  glab mr create [flags]

ALIASES
  new

FLAGS
      --allow-collaboration    Allow commits from other members
  -a, --assignee usernames     Assign merge request to people by their usernames
      --copy-issue-labels      Copy labels from issue to the merge request. Used with --related-issue
      --create-source-branch   Create source branch if it does not exist
  -d, --description string     Supply a description for merge request
      --draft                  Mark merge request as a draft
  -f, --fill                   Do not prompt for title/description and just use commit info
      --fill-commit-body       Fill description with each commit body when multiple commits. Can only be used with --fill
  -H, --head OWNER/REPO        Select another head repository using the OWNER/REPO or `GROUP/NAMESPACE/REPO` format or the project ID or full URL
  -l, --label strings          Add label by name. Multiple labels should be comma separated
  -m, --milestone string       The global ID or title of a milestone to assign
      --no-editor              Don't open editor to enter description. If set to true, uses prompt. (default false)
      --push                   Push committed changes after creating merge request. Make sure you have committed changes
      --recover                Save the options to a file if the merge request fails to be created. If the file exists, the options will be loaded from the recovery file (EXPERIMENTAL)
  -i, --related-issue string   Create merge request for an issue. The merge request title will be created from the issue if --title is not provided.
      --remove-source-branch   Remove Source Branch on merge
      --reviewer usernames     Request review from users by their usernames
  -s, --source-branch string   The Branch you are creating the merge request. Default is the current branch.
      --squash-before-merge    Squash commits into a single commit when merging
  -b, --target-branch string   The target or base branch into which you want your code merged
  -t, --title string           Supply a title for merge request
  -w, --web                    continue merge request creation on web browser
      --wip                    Mark merge request as a work in progress. Alternative to --draft
  -y, --yes                    Skip submission confirmation prompt, with --fill it skips all optional prompts
```

h3ndry:
[...]

PR #149, opened 2023-12-17
Feat: Create Merge Request

harrisoncramer:
This MR will allow you to create a merge request for the current branch without leaving Neovim.

This was previously deferred as a feature since the `glab` CLI tool already supports this, however, being able to do this inside of the plugin is core to resolving/merging/managing MRs.

This required a slight refactor of the Go server, which previously made a call to get the MR info at startup every time. Since the MR will not necessarily be created now at startup, we have moved that call into a middleware that's called prior to every route that requires the data, assuming that the `mergeId` property has not already been initialized in the client state.

The refactor therefore moves all routes that require the additional merge request call under the `/mr` route prefix.


harrisoncramer:
This MR will address #41

Issue #128, opened 2023-12-01
Create comment fails with error : empty line_code


aravind2612krishna:
## Bug Description
The create comment request returns the following response
``` {"message":"400 Bad request - Note {:line_code=>[\"can't be blank\", \"must be a valid line code\"]}"} ```
The request body is as follows:
```
{
  "body": "```suggestion:-4+0\n                redactedcode\n                    redactedcode-\u003eDestroy();\n\n                if (redactedcode)\n                    redactedcode-\u003eDestroy();\n```\n\nredactedcomment",
  "position": {
    "base_sha": "a2610248747681941279bd033169980fefd25732",
    "head_sha": "d08978d3d18fdfbd9945952d58c50507dc6a23ed",
    "start_sha": "a2610248747681941279bd033169980fefd25732",
    "new_path": "redactedfile",
    "old_path": "redactedfile",
    "position_type": "text",
    "new_line": 1659,
    "old_line": 1979,
    "line_range": {
      "start": {
        "line_code": "ecaaaac4a1a5cb79b55c2e7bf498f9e3847f87d0_1975_1655",
        "type": "new"
      },
      "end": {
        "line_code": "ecaaaac4a1a5cb79b55c2e7bf498f9e3847f87d0_1979_1659",
        "type": "new"
      }
    }
  }
}
```

## Reproduction Steps
call to create_comment(), create_multiline_comment() or create_comment_suggestion(),
It is not rare, some times it works.

johnybx:
Hi, 
I would need more information about this because I cannot reproduce it:
* was the end of selection outside of changed lines ( asking because should be when both `new_line` and `old_line` were specified )
* when you open gitlab UI and check the region you selected does the numbers match the request ? When you open MR in gitlab UI and point mouse over the line number you should see URL in form `/merge_requests/XYZ/diffs#line_code` which should match line_code in request.

harrisoncramer:
@johnybx I've noticed that you can reliably trigger this error in two different ways:

1. If you try to leave a comment on an unchanged line
2. If you try to leave a multi-line comment and the mutli-line comment spans across unchanged lines

aravind2612krishna:
> @johnybx I've noticed that you can reliably trigger this error in two different ways:
> 
>     1. If you try to leave a comment on an unchanged line
> 
>     2. If you try to leave a multi-line comment and the mutli-line comment spans across unchanged lines

[...]

PR #151, opened 2023-12-21
Fix: Discussion + Note Creation Bugs

harrisoncramer:
This MR is an attempt to address some of the bugs/issues with discussion creation and note creation.

This is related to #128 which is occurring due to the Gitlab API's _extremely particular_ API surrounding line codes. This certainly doesn't fix all the issues but it solves a few different ones.

First, when you left a multiline ranged comment we were grabbing the CURRENT line rather than the top of your visual selection and using that as the start line, which caused issues. 

Another bug was related to leaving a comment on an unchanged line in the _new_ pane. This is solved now -- we look at the current line, and also at the linked line in the other reviewer pane. If the two are identical, then you are actually leaving a comment on an old diff, so we adjust the payload correctly.

### DIFF
--- a/lua/gitlab/actions/discussions/signs.lua
+++ b/lua/gitlab/actions/discussions/signs.lua
@@ -192,7 +192,7 @@ M.parse_diagnostics_from_discussions = function(discussions)
         return {}, {}, string.format("Unsupported line range type found for discussion %s", discussion.id)
       end
     else -- Diagnostics for single line discussions.
-      if first_note.position.new_line ~= nil then
+      if first_note.position.new_line ~= nil and first_note.position.old_line == nil then
         local new_diagnostic = {
           lnum = first_note.position.new_line - 1,
         }
@@ -306,7 +306,7 @@ M.parse_signs_from_discussions = function(discussions)
       local sign = vim.tbl_deep_extend("force", {
         id = first_note.id,
       }, base_sign)
-      if first_note.position.new_line ~= nil then
+      if first_note.position.new_line ~= nil and first_note.position.old_line == nil then
         table.insert(new_signs, vim.tbl_deep_extend("force", { lnum = first_note.position.new_line }, sign))
       end
       if first_note.position.old_line ~= nil then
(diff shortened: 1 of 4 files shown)


lua/gitlab/actions/discussions/signs.lua
```diff
@@ -192,7 +192,7 @@ M.parse_diagnostics_from_discussions = function(discussions)
         return {}, {}, string.format("Unsupported line range type found for discussion %s", discussion.id)
       end
     else -- Diagnostics for single line discussions.
-      if first_note.position.new_line ~= nil then
+      if first_note.position.new_line ~= nil and first_note.position.old_line == nil then
```
harrisoncramer, line 195: Do not show diagnostics for comments that are on the old diff in the "b" pane

lua/gitlab/reviewer/diffview.lua
```diff
@@ -100,8 +105,11 @@ M.get_location = function(range)
     u.notify("Diffview reviewer must be initialized first")
     return
   end
+
   local bufnr = vim.api.nvim_get_current_buf()
-  local current_line = vim.api.nvim_win_get_cursor(0)[1]
+
+  -- If there's a range, use the start of the visual selection, not the current line
+  local current_line = range and range.start_line or vim.api.nvim_win_get_cursor(0)[1]
```
harrisoncramer, line 112: Use the top of the visual selection as the current line, not the bottom, when there is a range

lua/gitlab/reviewer/diffview.lua
```diff
@@ -120,7 +128,11 @@ M.get_location = function(range)
   local result = {}
   local type
   local is_new
-  if layout.a.file.bufnr == bufnr then
+
+  if
+    layout.a.file.bufnr == bufnr
+    or (M.lines_are_same(view.cur_layout) and layout.b.file.bufnr == bufnr and range == nil)
+  then
```
harrisoncramer, line 135: If leaving a comment in the "b" (aka new) pane, make sure that the line is not identical with the old line. If it is, then we are actually leaving a comment on the old diff.

PR #163, opened 2024-01-17
Feat: save popup contents to register

jakubbortlik:
In this small PR, I've added the possibility to save the contents of the popup window to a register. This is useful if `apply_action` fails and the user does not want to lose their content (e.g., a comment with a change suggestion).


briandipalma:
Should this not be enabled by default? I've had a few comments disappear and it was annoying.  Or maybe have it trigger on a submit action error if possible...

jakubbortlik:
I didn't make it the default behaviour because I didn't want to mess with people's registers. But I agree that it would be nice to have some safety mechanism that would prevent losing data when an error occurs and it is not handled by the register trick. That would be up to @harrisoncramer (or somebody else, @briandipalma ? :smile:) whether he wants to go deeper into this. I'm not that good with lua yet.

Issue #165, opened 2024-01-18
Discussion tree can be loaded into incorrect window


jakubbortlik:
## Bug Description

While Discussions data is being loaded, switching windows (e.g., to the reviewer, or the Diffview file panel) makes `gitlab.nvim` load the data in the newly focused window, not in the Discussions panel.

## Reproduction Steps

1. Check out the feature branch
2. Open Neovim
3. Start a review
4. Run `gitlab.toggle_discussions`
5. Press `<C-W><C-W>` while "Loading data..." is being displayed in the Discussions panel
6. Wait for the discussion tree to be loaded into the focused window

harrisoncramer:
Fixed by #167

PR #167, opened 2024-01-21
Fix: Write to correct window if user changes during loading state

harrisoncramer:


harrisoncramer:
This MR fixes issue #165 where you could accidentally load the discussions tree into the wrong window if you switched while the content was still loading.

Issue #164, opened 2024-01-18
Closing another diffview breaks gitlab.nvim's jumping to reviewer and making comments


jakubbortlik:
## Prerequsities

- [x] The "Troubleshooting" section of the README did not help
- [x] I've installed the required dependencies
- [x] I'm on the latest version of the plugin

## Bug Description

During code reviews I quite often use `DiffviewFileHistory` to see the full history of the file. However, closing the file history view triggers the `DiffviewViewClosed` user autocommand. This is then detected by `gitlab.nvim` (`lua/gitlab/reviewer/diffview.lua` line 33) and used to set the tab number to `nil` effectively disabling things like jumping to comments and determining the location within the diffview (and hence making comments).

I then have to restart `gitlab.nvim` or I just restart `nvim` (no big deal, but it would of course be nice if it could be avoided).

harrisoncramer:
Hey @jakubbortlik please check out the branch that I just created to make sure it works for you. I'm now checking that the closed tab # actually matches the tab number of the reviewer which should resolve this issue.

jakubbortlik:
Thanks for addressing the issues, I've been filing in recently. I'll try to have a look tomorrow morning.

jakubbortlik:
Hi, @harrisoncramer, I confirm that the fix in `165-discussion-tree-can-be-loaded-into-incorrect-window` works for me. Thanks!

And also the fix regarding loading the discussion tree in the wrong window works as expected. Thanks!

PR #168, opened 2024-01-21
Fix: De-Register Tabpage on Correct Close Event

harrisoncramer:
This MR attaches to Diffview's hooks API to get the actual tabpage of a closed Diffview, so that we only register the `gitlab.nvim` reviewer as being closed when the tab numbers line up. Previously we were closing it whenever this event fired which didn't work because someone could have both Diffview and Gitlab open at the same time.


harrisoncramer:
This MR addresses #164

PR #169, opened 2024-01-22
Fix: Allow Notes w/out Discussion Tree Open

harrisoncramer:
This MR fixes an issue where someone could leave a note on an MR without opening the discussion tree first. We want to avoid throwing an error by attempting to update the discussion tree if it doesn't already exist.

### DIFF
--- a/lua/gitlab/actions/discussions/init.lua
+++ b/lua/gitlab/actions/discussions/init.lua
@@ -429,6 +429,9 @@ local function nui_tree_prepare_node(node)
 end
 
 M.rebuild_discussion_tree = function()
+  if M.linked_bufnr == nil then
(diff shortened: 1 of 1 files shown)


lua/gitlab/actions/discussions/init.lua
```diff
@@ -442,6 +445,9 @@ M.rebuild_discussion_tree = function()
 end
 
 M.rebuild_unlinked_discussion_tree = function()
+  if M.unlinked_bufnr == nil then
+    return
+  end
```
harrisoncramer, line 450: Someone can leave a note before initializing the discussion tree, we want to avoid running this function if the tree hasn't been built yet.

PR #174, opened 2024-01-30
Feat: Make MR title input window configurable

jakubbortlik:
This mini PR makes it possible for users to configure the MR title input window.

It seems that "position" cannot be configured, but "size" is the most important anyway, to make writing long MR titles easier. Setting "border" is enabled just for UI consistency.

### DIFF
--- a/lua/gitlab/actions/create_mr.lua
+++ b/lua/gitlab/actions/create_mr.lua
@@ -78,21 +78,6 @@ local description_popup_settings = {
   },
 }
 
----Create input options after user settings have been merged with defaults.
-local get_title_input_options = function()
-  return {
-    position = "50%",
-    relative = "editor",
-    size = state.settings.create_mr.title_input.width,
-    border = {
-      style = state.settings.create_mr.title_input.border,
-      text = {
-        top = "Title",
-      },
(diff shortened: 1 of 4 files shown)


lua/gitlab/actions/create_mr.lua
```diff
@@ -78,17 +78,20 @@ local description_popup_settings = {
...
+  return {
+    position = "50%",
+    relative = "editor",
+    size = state.settings.create_mr.title_input.width,
+    border = {
+      style = state.settings.create_mr.title_input.border,
+      text = {
+        top = "Title",
+      },
     },
```
harrisoncramer, line 92: Would prefer to just pass this state directly inside of the `add_title` function, think it's a little nicer to not have to split this up:

Screenshot 2024-02-10 at 11 17 14 AM

harrisoncramer:
Made one small suggestion, otherwise this looks good to me

lua/gitlab/actions/create_mr.lua
jakubbortlik: @harrisoncramer Thanks for the review. I've applied your suggestion.

PR #176, opened 2024-02-07
Feat collapse and expand nodes

jakubbortlik:
Hi @harrisoncramer, I'd like to ask you to have a look at this MR in which I've tried to implement methods for collapsing and expanding discussion nodes. I find this functionality mainly helpful for more easily searching the discussion tree and then closing selected nodes.

I've added some default mappings for:
- collapsing all nodes
- collapsing all nodes except the current one (open the current one if it was closed before)
- collapsing all nodes except the unresolved discussions
- expanding all nodes
- expanding all unresolved nodes (collapses resolved discussions)

I'll be glad for any feedback and suggestions for improvement.

Besides, this MR contains a couple documentation fixes and a small adjustment to the `help` popup.


lua/gitlab/actions/help.lua
```diff
@@ -10,7 +10,7 @@ M.open = function()
   local help_content_lines = {}
   for _, keymap in ipairs(keymaps) do
     if keymap.desc ~= nil then
-      local new_line = string.format("%s: %s", keymap.lhs, keymap.desc)
+      local new_line = string.format("%s: %s", keymap.lhs:gsub(" ", "<space>"), keymap.desc)
```
harrisoncramer, line 13: Thanks, good catch

harrisoncramer:
This is awesome, and works just as you've described. However, I'd make two suggestions:

1. Having to deal with different keybindings for expanding/collapsing resolved/unresolved discussions is a bit of a headache for the user. It's easier in my opinion to have a universal "toggle" command for both collapsing and expanding a particular type of node. We could then condense the bindings down to:

- TR (Toggle resolved)
- TU (Toggle unresolved)
- TT (Toggle all)

The core of your MR wouldn't have to change much at all, you could just track some state (probably a boolean) about whether those types of nodes are expanded or collapsed and then do the opposite. For the "all" toggle we'd set both to true/false depending on their current state.

2. I think that it makes sense to use `T` and just have this be a breaking change (since we use `T` right now to switch windows) in order to have it be consistent with `t` which is toggle the current node.

harrisoncramer:
Oh and while we're breaking keybindings, let's change the help popup to `g?` instead of just `?` like you suggested earlier so that folks can query up/down using normal vim commands.

jakubbortlik:
Hi @harrisoncramer, thanks for the review and for your suggestions. I'm sorry that the following is so lengthy, I just want to make sure, I understand how you'd like me to change the MR and I explain a few points why I did it the way I did.

I used different keybindings for collapsing and expanding nodes mainly because I thought about it more like closing and opening folds - those too have different keybindings, `zc` and `zo`, respectively. To me it made sense since it always gives you the expected result - you want to close everything, you just close it. Whereas with toggling, you can have a mixture of open and closed nodes and then you can forget whether your toggle is on or off and then you may have to press the keybinding twice to get the desired result. This is not a big deal though, after I thought about your suggestion for a moment, I realized that I will anyway mostly just want to open all nodes, search for a string and then close all nodes except the one with the match I want. I believe this would work with toggle keybindings too, only I'd have to press "t" after closing all nodes, or TT would just have to leave the current node open (this was the behaviour of my "CO" mapping).

Just to make sure I understand your idea about the toggling correctly:
1. There would be two Booleans, e.g., `show_resolved`, and `show_unresolved`
2. All nodes will be closed at first, `show_resolved = false`, and `show_unresolved = false`
3. If I press TR `show_resolved` becomes `true` and all resolved discussions will be opened. What should happen to any manually opened unresolved discussions - should they get closed?
4. If I then press TT - "toggle all", `show_resolved` becomes `false` again, hiding resolved discussions, and `show_unresolved` becomes `true`, and unresolved discussions would be shown. What happens to the current discussion? Should it always be shown even if it is in the "wrong" state, i.e., resolved, in this example? I'm not sure this behaviour is very practical.
5. If I wanted to actually open *all* discussions, then I'd have to think about which discussions are closed at the moment and then select the appropriate mapping - I don't like this idea very much because I prefer not to think about my mappings too much :)
6. Alternatively to point 4, `TT` could work like this: if any of the `show_resolved` and `show_unresolved` is `false` then set **both** to `true`, on the next `TT` set both to `false`.

What do you think the desired behaviour is? Also importantly, do you expect to just use a single method `toggle_nodes` instead of the two `expand_nodes` and `collapse_nodes`?

What do you think about using just "R", "U" and "T" - without the "T" prefix (yeah, I'm lazy when it comes to typing)? Especially "TU" is a little cumbersome on a qwerty keyboard. The question is whether you want to reserve "R" and "U" for some other possible functionality in the future. There are still some quite a few keys available since the buffer is non-modifiable: "a, A, c, C, d, D, i, I, J, m, O, P, s, S, x, X" (and some others, probably less practical: ">, <, =, ~, !, &, ., g&, g?, gI, gJ, gP, gQ, gR, gU, gp, gq, gr, gw").

What should be the mapping to replace the current "T" for "switch view type"? I suggest "S" as in "**S**witch" :) I wouldn't want to use "V" because linewise visual mode is too useful and "S" is useless in a non-modifiable buffer anyway (except maybe some people use "S" for some other plugins, such as [Flash](https://github.com/folke/flash.nvim)).

After trying to sum it up, I still feel inclined to using separate keybindings for closing and opening nodes, except now, I wouldn't use a separate "CO" binding, but I'd always leave the current discussion open, even with "CA", and more often than not, I'd just be using "EA" and "CA", anyway.

So here are some more questions:
1. Would you be OK with having both toggling and non-toggling mappings (and the non-toggling methods in the code)?
2. Do you think it would be possible to make the methods and their configuration available to the user? I think this might be useful if somebody would like to creating their own toggling rules (e.g., toggle all nodes that contain some specific text, user name, etc.). I don't see how this could be easily done, since the methods accept the `tree` as an argument.

jakubbortlik:
What currently does not work as I'd expect is collapsing all nodes except the current one (`M.collapse_nodes(tree, { collapse_unresolved = true, keep_current_open = true })`) when triggered on a `file_name` or `path` node in the `by_file_name` tree type. This just expands *all* nodes in the tree, whereas I'd either expect it to expand maybe all discussions for that file/path or just collapse everything. Maybe this will be fixed when I rewrite the way how the `opts` are used.

harrisoncramer:
You've asked a lot of different questions, so it's a bit difficult to know what exactly you want me to answer.

The ability to collapse and expand nodes in the tree seems to be a pretty straightforward feature, let's not overcomplicate this. 

I'd imagine that most users will likely just want to just expand all the nodes, search for something and then do some action, possibly followed by collapsing other discussions. As long as your solution makes that possible then I'm fine with your additional functionality; my original suggestion was just a way to possibly make things simpler. Keeping the collapse and expand key bindings separate is fine.

Exposing a customizable function for bulk actions against the tree sounds like it could be a good idea as well, but I don't think you need to try to tackle that in this MR. 

I'd try to limit the scope of your changes here to achieving the collapsing and expanding. If we wanted to follow up with bulk actions, I would imagine that it would take some sort of "filter" option, which would indicate which notes the actions should be applied to, and then a callback function, which would accept the node as an argument in order to do some manipulation.

Please let me know when this is ready for another review and I'll go through the code.

jakubbortlik:
Hi @harrisoncramer,
thanks for your feedback - you are right, I made it too complicated. Now I've removed my previous code, implemented the toggling method and made it just enough customizable so that I get from it what I need, and I think it's now really clearer for the user, as you said. Can you please review the PR again?

In the end I've set these simple default keymaps: T - "toggle all", R - "toggle resolved", U - "toggle unresolved", S - "Switch view". Let me know if you'd rather go with the prefixed versions TT, TR, TU.

README.md
```diff
@@ -136,6 +136,10 @@ require("gitlab").setup({
     delete_comment = "dd", -- Delete comment
     reply = "r", -- Reply to comment
     toggle_node = "t", -- Opens or closes the discussion
+    toggle_all_discussions = "T", -- Open or close separately both resolved and unresolved discussions
+    toggle_resolved_discussions = "R", -- Open or close all resolved discussions
+    toggle_unresolved_discussions = "T", -- Open or close all unresolved discussions
```
harrisoncramer, line 141: This should be "U" I think

harrisoncramer:
This works beautifully, nice job dude! I left one tiny comment on the readme but once that's fixed I'll merge this in. Thank you for the awesome contribution, I'll definitely be using this feature all the time!

README.md
jakubbortlik: @harrisoncramer Sure, thanks for the catch. I was changing it back and forth in three different places and this one I missed :) It's fixed now.

PR #180, opened 2024-02-13
Fix: Line Comment Refactor

harrisoncramer:
This MR overhauls the logic that we previously had in place for determining payloads for creating comments.

Gitlab's API is notoriously awful when it comes to posting comments, because the actual content of the POST has to change depending on whether or not you are sending a comment on a modified line, an unchanged line, or a deleted line. 

This MR takes a very meticulous approach to determining what sort of change has occurred on the currently focused line. It does this by parsing the hunk headers and _actual content of the diff_ as well, to figure out whether the line has been added, changed, or untouched. It also takes advantage of Neovim's scroll lock functionality. If you are scrolling in the reviewer panel, the location of your cursor in the changed file and in the new file will be "locked" together. This let's us send not only the line in the current buffer (if you are in the new panel) but also the line number from the old buffer as well.

This MR is meant to resolve all issues with leaving single-line comments, as outlined in issue #128 

The plan is to keep this feature in a feature branch for some time and to have some users of the application make sure that it's working as intended. Once it's deemed stable enough it'll be merged into main.

THIS MR DOES NOT ATTEMPT TO ADDRESS OTHER ISSUES WITH MULTI-LINE COMMENTS.

There could still be issues with multi-line comments where the range is off for some reason. This MR will only fix all isues related to line codes being wrong.

### DIFF
--- a/lua/gitlab/reviewer/diffview.lua
+++ b/lua/gitlab/reviewer/diffview.lua
@@ -197,7 +197,7 @@ M.get_location = function(range)
     -- The line was not found in any hunks, only send the old line number
   elseif modification_type == "unmodified" or modification_type == "bad_file_unmodified" then
     reviewer_info.old_line = a_linenr
-    reviewer_info.new_line = a_linenr
+    reviewer_info.new_line = b_linenr
   end
 
   if range == nil then
@@ -339,17 +339,22 @@ end
 ---@param hunks Hunk[] A list of hunks
 ---@param all_diff_output table The raw diff output
 function M.get_modification_type(a_linenr, b_linenr, is_current_sha, hunks, all_diff_output)
-  -- If leaving a comment on the old window, we can either be commenting
-  -- on a deletion or an unmodified line. It's a deletion if it's in the range
-  -- of the hunks and the new range is zero, since that is only a deletion hunk,
-  -- or if we find a match in another hunk with a range, then we should check if that line
-  -- was actually deleted. To do that we have to see whether that line is prefixed
-  -- with a "-" only. If it is, then it's a deletion.
-  if not is_current_sha then
-    for _, hunk in ipairs(hunks) do
-      local old_line_end = hunk.old_line + hunk.old_range
-      local new_line_end = hunk.old_line + hunk.new_range
-      if a_linenr >= hunk.old_line and a_linenr <= old_line_end and hunk.new_range == 0 then
+  for _, hunk in ipairs(hunks) do
+    local old_line_end = hunk.old_line + hunk.old_range
+    local new_line_end = hunk.old_line + hunk.new_range
+
+    -- If leaving a comment on the old window, we can either be commenting
+    -- on a deletion or an unmodified line.
+    if is_current_sha then
+      if b_linenr >= hunk.new_line and b_linenr <= new_line_end - 1 then
+        return "added"
+      end
+    else
+      -- It's a deletion if it's in the range of the hunks and the new
+      -- range is zero, since that is only a deletion hunk, or if we find
+      -- a match in another hunk with a range, and the corresponding line is prefixed
+      -- with a "-" only. If it is, then it's a deletion.
+      if a_linenr >= hunk.old_line and a_linenr <= old_line_end and hunk.old_range == 0 then
         return "deleted"
       end
       if (a_linenr >= hunk.old_line and a_linenr <= old_line_end) or (a_linenr >= hunk.new_line and b_linenr <= new_line_end) then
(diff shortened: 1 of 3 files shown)


harrisoncramer:
@jakubbortlik Hey would you be interested in using this branch for a while to see if it resolves the issues with incorrect line codes? I'm going to use it for the next week or so but the more people using it the better. Once your toggle branch merges into main I'll bring that into this MR as well, so it should have all the features you want.

The goal for now is to fix single-line comment/note creation issues completely. Once you are reliably able to create single-line comments without any issues, we can start to address the other issues, where those line codes become out of date with the current diff, which is a separate issue.

lua/gitlab/reviewer/diffview.lua
```diff
@@ -126,106 +126,117 @@ M.get_location = function(range)
...
     return
   end
 
-  local current_line_info
-  if is_new then
-    current_line_info = u.get_lines_from_hunks(hunks, result.new_line, is_new)
-  else
-    current_line_info = u.get_lines_from_hunks(hunks, result.old_line, is_new)
+  local a_linenr = vim.api.nvim_win_get_cursor(a_win)[1]
+  local b_linenr = vim.api.nvim_win_get_cursor(b_win)[1]
```
harrisoncramer, line 173: Here we take advantage of scroll-lock to get the line from the current and old buffers. We can use the old line even if we're in the new buffer this way.

lua/gitlab/reviewer/diffview.lua
```diff
@@ -126,106 +126,117 @@ M.get_location = function(range)
...
+    reviewer_info.old_line = nil
+    reviewer_info.new_line = b_linenr
+    -- Comment on deleted line: Include only new_line in payload.
+  elseif modification_type == "deleted" then
+    reviewer_info.old_line = a_linenr
+    reviewer_info.new_line = nil
+    -- The line was not found in any hunks, only send the old line number
+  elseif modification_type == "unmodified" or modification_type == "bad_file_unmodified" then
+    reviewer_info.old_line = a_linenr
+    reviewer_info.new_line = b_linenr
```
harrisoncramer, line 201: Here we're actually building the payload correctly depending on what the helper `get_modification_type` returns.

jakubbortlik:
> @jakubbortlik Hey would you be interested in using this branch for a while to see if it resolves the issues with incorrect line codes? I'm going to use it for the next week or so but the more people using it the better. Once your toggle branch merges into main I'll bring that into this MR as well, so it should have all the features you want.
> 
> The goal for now is to fix single-line comment/note creation issues completely. Once you are reliably able to create single-line comments without any issues, we can start to address the other issues, where those line codes become out of date with the current diff, which is a separate issue.

I'm exited about this! Of course I'll test the branch. I've been filing some fun UI feature requests recently, but creating comments successfully is of course a priority, I just didn't know how I could help other than providing logs and descriptions of cases that don't work.

Just a clarification - this only addresses the comments created with `gitlab.create_comment` in `normal` mode not the `gitlab.create_multiline_comment` in `virtual` mode even if the selection contained just one line, right? I guess from the point of `gitlab.nvim` this would still be a "multiline" comment.

harrisoncramer:
Yes that's right. I'm going to try and track down any issues with multi-line comments later. This also doesn't address comments being in the wrong place after an update, etc. -- just the initial POST.

lua/gitlab/utils/init.lua
```diff
@@ -515,12 +524,37 @@ end
...
+  if line:sub(1, 2) == "@@" then
+    -- match:
+    --  @@ -23 +23 @@ ...
+    --  @@ -23,0 +23 @@ ...
+    --  @@ -41,0 +42,4 @@ ...
+    local old_start, old_range, new_start, new_range = line:match("@@+ %-(%d+),?(%d*) %+(%d+),?(%d*) @@+")
+
+    return {
+      old_line = tonumber(old_start),
+      old_range = tonumber(old_range) or 0,
```
jakubbortlik, line 544: Hi @harrisoncramer, first of all, thank you for all your work on this - the more I see of the code of `gitlab.nvim` the more I admire what you've achieved and given to the Neovim community.

I think this line should be changed like this:
```suggestion
      old_range = tonumber(old_range) or 1,
```
Here's why: I've been trying to break the new way one-line comments are created and it seems I've found at least one way. In branch `testing1` I've got `file1.txt` with the following content:
```
abc
```
In branch `testing2`, the content of `file1.txt` is changed to this:
```
def
ghi
```
I'm trying to merge `testing2` into `testing1`. The diff hunk looks like this:
```
diff --git a/file.txt b/file.txt
index 8baef1b..aff48dc 100644
--- a/file.txt
+++ b/file.txt
@@ -1 +1,2 @@
-abc
+def
+ghi
```
Now, if I try to put a comment on line 1 in the "new" window ("def"), I'm getting the following error:
```
gitlab.nvim: Could not create discussion: POST https://gitlab.cloud.phonexia.com//api/v4/projects/459/merge_requests/4/discussions: 400 {message: 400 Bad request - Note {:line_code=>["can't be blank", "must be a valid line code
"]}}
```
I've tried to debug this and the problem seems to be that if I delete one line and replace it with two (or more) lines, `old_range` is calculated as `0` and in the loop starting at line 418 in `lua/gitlab/reviewer/diffview.lua`, the `hunk_line_index` is off by `1`, and the "def" is not recognized as `"added"`, because what we are actually matching against `"^%+"` is `"-abc"` and the resulting `modification_type` is determined as `"bad_file_unmodified"`. The following addition `"ghi"` is then recognized as `"added"`, because `"^%+"` is matched against `"+def"`.

I'm not sure if `old_range = tonumber(old_range) or 1` doesn't break anything else, but it certainly fixes the error I just described, and I believe you should be able to reproduce easily.

harrisoncramer: Hey, thanks for catching this.

Your solution worked well but I needed to account for one other edge case. I pushed up that fix. We only want to set that number to 1 when the old range is zero AND the old line is zero as well. I think this fixes everything...

aravind2612krishna: Thanks @jakubbortlik and @harrisoncramer for fixing this.

PR #181, opened 2024-02-14
Add/Show Emojis on Notes

harrisoncramer:
Adds the ability to view emojis on a discussion.


harrisoncramer:
The backend implementation for this is relatively straightforward to fetch Emojis, however some are Unicode characters that do not display correctly in terminals.

The complex part of this is translating those shortcodes (e.g. :thumbs_up: to 👍 ) into actual emojis, detecting when an invalid emoji has been created in Gitlab, and deciding how we want to handle it in the terminal. I'm thinking that if we detect an invalid one we should just show the literal shortcode.

jakubbortlik:
Actually, I can see the 👍 icon in my terminal just fine:
image
I'm using the [foot](https://github.com/DanteAlighierin/foot) terminal. I've also tried it in Alacritty and it works. At first I thought this character was provided by the [nvim-tree/nvim-web-devicons](https://github.com/nvim-tree/nvim-web-devicons) plugin, but I can see it even after uninstalling the plugin.

But sure, there should be a way that would work even for terminals that do not support such characters. Is there a way I can help with this?

harrisoncramer:
That emoji is a single Unicode character. You'll notice that emojis like 👋🏾 or 👩‍🚒 are not, and will be displayed incorrectly, as two or three separate emojis in sequence because terminals do not support showing skin tone modifiers or composite emojis.

harrisoncramer:
No need to help I'll figure something out. Thanks though.

jakubbortlik:
> That emoji is a single Unicode character. You'll notice that emojis like 👋🏾 or 👩‍🚒 are not, and will be displayed incorrectly, as two or three separate emojis in sequence because terminals do not support showing skin tone modifiers or composite emojis.

OK, I didn't realize this. It's funny though, how the firewoman emoji is decomposed into this:
image

harrisoncramer:
I'm also not going to address the issue with multi-glyph emojis, as these are not currently supported by Alacritty. This is not an issue with this plugin but should be dealt with at the presentation layer by the end-user, as this will be broken for any other plugins that attempt to use them.

This MR is now done, I'll use it for the next week or so to iron out any issues and then merge into main!

Issue #182, opened 2024-02-14
discussions tree does not contain all discussions


jakubbortlik:
## Prerequsities

- [x] The "Troubleshooting" section of the README did not help
- [x] I've installed the required dependencies
- [x] I'm on the latest version of the plugin

## Bug Description

- The discussions tree does not contain all the discussions that are shown in Gitlab in the browser.
- The discussions are missing in the log (gitlab.nvim.log)
- The strange thing is, that I can see some of the discussions, I can successfully do things like `curl --header "PRIVATE-TOKEN: ${TOKEN}" localhost:44859/mr/info` and see information about the MR.
- I can even create new comments in nvim and they show up in Gitlab!
- The `discussions_tree.blacklist` is empty, but that should probably play no when the data is not in the log in the first place.

But I see the following in the log, notice the `page=1` in all three requests. Indeed if I replace it with `page=2`, and paste it in the browser, I can see the missing discussions!

```
-- REQUEST --
GET //api/v4/projects/42/merge_requests/296/discussions?page=1&per_page=250 HTTP/1.1
Host: _____
Accept: application/json
Private-Token: REDACTED
User-Agent: go-gitlab

-- REQUEST --
GET //api/v4/projects/42/merge_requests/296/discussions?page=1&per_page=250 HTTP/1.1
Host: _________
Accept: application/json
Private-Token: REDACTED
User-Agent: go-gitlab

-- REQUEST --
GET //api/v4/projects/42/merge_requests/296/discussions?page=1&per_page=250 HTTP/1.1
Host: __________
Accept: application/json
Private-Token: REDACTED
User-Agent: go-gitlab
```

And also this:
```
-- RESPONSE --
HTTP/2.0 200 OK
Content-Length: 272544
Cache-Control: max-age=0, private, must-revalidate
Content-Type: application/json
Date: Wed, 14 Feb 2024 16:15:54 GMT
Etag: W/"02da1db492c0b81a67fac76d4298ceb3"
Link: <__________/42/merge_requests/296/discussions?id=42&noteable_id=296&page=2&per_page=100>; rel="next", <__________/42/merge_requests/296/discussions?id=42&noteable_id=296&page=1&per_page=100>; rel="first", <_____________/42/merge_requests/296/discussions?id=42&noteable_id=296&page=2&per_page=100>; rel="last"
Referrer-Policy: strict-origin-when-cross-origin
Strict-Transport-Security: max-age=63072000
Vary: Origin
[...]

PR #183, opened 2024-02-14
Fix: Iterate over several pages to get all discussions

jakubbortlik:
This is a naive fix for #182. I've just added a loop that goes over several pages of discussions to get more items than the default 100 allowed by Gitlab. `gitlab.nvim` tries to get 250, but cannot get more than 100 in one request.

### DIFF
--- a/cmd/list_discussions.go
+++ b/cmd/list_discussions.go
@@ -62,42 +62,49 @@ func (a *api) listDiscussionsHandler(w http.ResponseWriter, r *http.Request) {
 		handleError(w, err, "Could not unmarshal request body", http.StatusBadRequest)
 	}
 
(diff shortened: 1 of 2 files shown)


jakubbortlik:
Hi, @harrisoncramer, I'd like to ask for help in fixing the tests for this. I don't know how to do that, this is the first time I've read and written some `go` code :) I've had a look at the [output](https://github.com/harrisoncramer/gitlab.nvim/actions/runs/7908365782/job/21587396213?pr=183) of the test, but cannot make anything out of it.

jakubbortlik:
Closing, as the simple suggested solution conflicts with the newly added code for emojis and there'll certainly be a better way to fix the problem with unlisted discussions.

PR #192, opened 2024-02-24
Feat: Improve discussion tree toggling

jakubbortlik:
This PR tries to make toggling of the discussion tree a little more user friendly. It will:
- Enable toggling nodes from the note body
- Enable toggling resolved status from child nodes
- Add keymap for toggling tree type

Apart from that, it fixes a small bug which caused an error whenever the cursor moved to a `file_name` or `path` node in the "by_file_name" discussion tree.


harrisoncramer:
I'm changing the way releases are handled to reduce release churn and to allow myself a non-stable development branch. Please change the target of your MR to `develop` instead of `main` which will aggregate multiple patches + such prior to release.

Issue #222, opened 2024-03-14
allow Squash and RemoveSourceBranch in CreateMrRequest


jakubbortlik:
## Feature Description

Currently it's possible to squash commits and request the source branch to be removed when _accepting_ a MR. I'd like to be able to specify these options already when _creating_ a MR, so that I don't have to go to the website to do this.

jakubbortlik:
Hi @harrisoncramer It seems that adding the parameters should be fairly straightforward. I'll be glad to submit a PR if you agree.

harrisoncramer:
Yes, this would be a nice addition. Currently we do not support this at time of merging (glM) or creating (glO) an MR, or make it visible to the user in the MR summary. I'd be open to an MR that implements this. Might be worth considering adding settings that would allow the user to specify default behaviors for this as well.

Issue #219, opened 2024-03-11
Mappings to change focus


benedikt-bartscher:
Provide mappings to change focus f.e. from MR Title to Description using apis like `nvim_set_current_win(title_popup.winid)`.

benedikt-bartscher:
Thank you, @jakubbortlik! I just tried `<c-j>` and `<c-k>` which did not work for some reason.

jakubbortlik:
Hi @benedikt-bartscher if the builtin keymaps do what you need, can you please close this issue?

benedikt-bartscher:
Hi @jakubbortlik they indeed work, however I would love to add support for nvims other window keymaps aswell.

jakubbortlik:
HI @benedikt-bartscher, I think this request would be more appropriate at the [nui.nvim](https://github.com/MunifTanjim/nui.nvim) library. There already exists a closed [discussion](https://github.com/MunifTanjim/nui.nvim/discussions/289) with basically the same topic, but it was more of an inquiry rather than a feature request, and it was asked from the point of view of a plugin author.

"nvims other window keymaps" is rather broad. Is there something in particular that you'd like to be able to do in the MR summary popup? I don't see much use myself in supporting `<ctrl-w>j` and `<ctrl-w>k` when `<ctrl-w><ctrl-w>` and `<ctrl-w>W` do the job as well. But I can see that e.g., using the `<tab>` key to cycle through the windows in the summary would be practical (something like the [resolution](https://github.com/RobinTruax/resolution/blob/63cbc5a50eee8df039f54e593ed377f04e60e583/lua/computation/napkin/actions.lua#L63) plugin does). Maybe if you are more specific in your feature request, Harrison might reopen it. Or you could create a PR yourself, if you are interested.

PR #230, opened 2024-04-02
Feat add squash and removesourcebranch to createmrrequest

jakubbortlik:
This MR addresses issue #222 by:
1. Showing the `delete_branch` and `squash` information in the MR Summary.
2. Enabling setting these options in the Create MR popup
3. Enabling setting default values for these options in `settings.create_mr`
4. Adding the necessary options in `cmd/create_mr.go`

It also makes some minor additional changes:
1. Cleanup of docs (especially the `gitlab.create_mr` parameters)..
2. Some refactoring of Box popup settings to make adding new options easier.
3. Some refactoring of the `pick_template` and `pick_target` functions to make adding new options easier.

I think some other changes should be implemented in order to improve the user experience but they are not part of this MR:
1. The behaviour of `gitlab.merge()` should be adjusted. Now the equally named options of `gitlab.merge()` simply override the options set by `gitlab.create_mr()` or selected in the browser (possibly by another user). I believe, by default, `merge()` should respect the settings that can be seen in the "Summary" (either set by `gitlab.create_mr()` or in the browser). I can imagine a situation where I wanted to override these, so maybe `gitlab.merge()` could still be passed arguments when called from the command line or through a mapping, but I would suggest to remove the default `settings.merge.squash` and `settings.merge.delete_branch` options, however, this would be a breaking change.
2. The confirmation popup now consists of 5 individual popups. Navigation with `<ctrl-w><ctrl-w>` and `<ctrl-w>W` is now more cumbersome than before, so I think it would be nice to implement some cycling mappings requested in the rejected issue #219. Then it would be easy to also use these mappings for the Summary popup.
3. Create `linewise_actions` that would enable changing the target branch and `delete_branch` and `squash` options in the confirmation popup, and also some select values in the Details window in the MR Summary. For the confirmation popup this is not that crucial -- a user can start over again if they make a mistake, but as soon as a MR is created, changes can only be made to it in the Browser, so making it possible to adjust some values in the Summary popup would be very practical.
4. Remove the current branch from the possible merge targets in `create_mr.pick_target()` - this could be done globally by adding a parameter to the following function, something like: `utils.get_all_git_branches(remote, include_current_branch)` function. Otherwise an error is thrown only after all the MR creation work has been done.
5. Update the MR Summary after values in the `Details` change (e.g., reviewers, assignees, status, etc.)

@harrisoncramer, if you think any of these additions make sense I can open feature requests for them, or separate MRs directly.


harrisoncramer:
Thank you for writing this MR! It looks very nice, I'll take a look this weekend

lua/gitlab/actions/create_mr.lua
```diff
@@ -224,22 +188,42 @@ M.open_confirmation_popup = function(mr)
   local function exit()
     local title = vim.fn.trim(u.get_buffer_text(M.title_bufnr))
     local description = u.get_buffer_text(M.description_bufnr)
-    local target = vim.fn.trim(u.get_buffer_text(target_popup.bufnr))
+    local target = vim.fn.trim(u.get_buffer_text(M.target_bufnr))
```
jakubbortlik, line 191: Changes like this were not really necessary, since both `target_popup.bufnr` and `M.target_bufnr` are the same number, but I thought it makes the code easier to read if buffer numbers are accessed in the same way.

lua/gitlab/actions/create_mr.lua
```diff
@@ -288,19 +278,27 @@ M.create_mr = function()
...
   local internal_layout
   internal_layout = Layout.Box({
     Layout.Box({
       Layout.Box(title_popup, { grow = 1 }),
       Layout.Box(target_branch_popup, { grow = 1 }),
     }, { size = 3 }),
+    Layout.Box({
+      Layout.Box(delete_branch_popup, { grow = 1 }),
+      Layout.Box(squash_popup, { grow = 1 }),
+    }, { size = 3 }),
```
jakubbortlik, line 301: I also experimented with a different layout where all the small popups (title, branch, delete, squash) would be on a single line with different popup widths, but I thought this might not work well on a very narrow (or split) screen.

harrisoncramer: I'm inclined to actually put the metadata fields _below_ the description, like they are on Gitlab. If you shorten the titles a bit they can fit easily, even on smaller screen sizes: 

Screenshot 2024-04-04 at 7 17 03 PM

This is achieved like so: 

```lua
  internal_layout = Layout.Box({
    Layout.Box({
      Layout.Box(title_popup, { grow = 1 }),
    }, { size = 3 }),
    Layout.Box(description_popup, { grow = 1 }),
    Layout.Box({
      Layout.Box(target_branch_popup, { grow = 1 }),
      Layout.Box(delete_branch_popup, { grow = 1 }),
      Layout.Box(squash_popup, { grow = 1 }),
    }, { size = 3 }),
  }, { dir = "col" })
```

harrisoncramer (changes requested):
This looks great, left a couple of small suggestions.

The other thing that we have to do as part of this MR is remove the `squash_message` option which is currently present. I think it's just a matter of addressing these three references, since this is now baked into the MR itself:

Screenshot 2024-04-04 at 7 29 59 PM

lua/gitlab/actions/create_mr.lua
jakubbortlik: Oh, nice! I like the metadata below the description. Since "Delete source branch" and "Squash commits" don't need to grow, I'd put them to the left with a fixed width and let the "Target branch" take up the rest of the window, like this:
image
And truncate the titles when the window length is below some limit, like this:
image

This is fairly easy to do:
```lua
  local delete_title = vim.fn.winwidth(0) > 110 and "Delete source branch" or "Delete source"
  local delete_branch_popup = Popup(u.create_box_popup_state(delete_title, false))
  M.delete_branch_bufnr = delete_branch_popup.bufnr
  local squash_title = vim.fn.winwidth(0) > 110 and "Squash commits" or "Squash"
  local squash_popup = Popup(u.create_box_popup_state(squash_title, false))
  M.squash_bufnr = squash_popup.bufnr

  local internal_layout
  internal_layout = Layout.Box({
    Layout.Box({
      Layout.Box(title_popup, { grow = 1 }),
    }, { size = 3 }),
    Layout.Box(description_popup, { grow = 1 }),
    Layout.Box({
      Layout.Box(delete_branch_popup, { size = {width = #delete_title + 4} }),
      Layout.Box(squash_popup, { size = {width = #squash_title + 4} }),
      Layout.Box(target_branch_popup, { grow = 1 }),
    }, { size = 3 }),
  }, { dir = "col" })
```

lua/gitlab/actions/create_mr.lua
```diff
@@ -224,22 +188,42 @@ M.open_confirmation_popup = function(mr)
...
   end
 
   local description_lines = mr.description and M.build_description_lines(mr.description) or { "" }
 
+  local delete_branch
+  if mr.delete_branch ~= nil then
+    delete_branch = mr.delete_branch
+  else
+    delete_branch = state.settings.create_mr.delete_branch
+  end
```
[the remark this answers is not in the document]
jakubbortlik: I actually tried the way you suggest before, but the problem is that it does not work for Boolean values. If `mr.delete_branch` is `false` it's not able to override the default value `state.settings.create_mr.delete_branch` if that is `true`.

How about extracting the logic to a utils function like this:
```lua
---Get the first non-nil value in the input table or nil
---@param values table The list of input values
---@return any
M.get_first_non_nil_value = function(values)
  for _, val in pairs(values) do
    if val ~= nil then
      return val
    end
  end
end
```
Then I'll be able to simplify my code to:
```lua
  local delete_branch = u.get_first_non_nil_value({mr.delete_branch, state.settings.create_mr.delete_branch})
```
Which is quite explicit about what it does.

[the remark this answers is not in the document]
jakubbortlik: By the way, I wanted to add tests for this function in `tests/spec/util_spec.lua`, but when running `busted tests/spec/util_spec.lua` I get the following error, followed by all the errors resulting from `gitlab.utils` not being loaded properly:
```
Failure → tests/spec/util_spec.lua @ 2                                                                                 
utils/init.lua Loads package                                                                                           
tests/spec/util_spec.lua:4: Expected objects to be the same.                                                           
Passed in:                                                                                                             
(boolean) false                                                                                                        
Expected:                                                                                                              
(boolean) true
```
Could you please add information in `.github/CONTRIBUTING.md` about how to run the tests?

jakubbortlik:
> This looks great, left a couple of small suggestions.
> 
> The other thing that we have to do as part of this MR is remove the `squash_message` option which is currently present. I think it's just a matter of addressing these three references, since this is now baked into the MR itself:
> 
> Screenshot 2024-04-04 at 7 29 59 PM

I don't think this is correct. The `state.settings.popup.squash_message` setting is for customizing the squash message popup and I think we need to keep that. `state.settings.merge.squash` on the other hand, defines whether commits should be squashed, but I think we should get rid of the whole `state.settings.merge` option, like this:

```diff
diff --git a/README.md b/README.md
index 5801d69..b41ac71 100644
--- a/README.md
+++ b/README.md
@@ -201,10 +201,6 @@ require("gitlab").setup({
     success = "✓",
     failed = "<U+F467>",
   },
-  merge = { -- The default behaviors when merging an MR, see "Merging an MR"
-    squash = false,
-    delete_branch = false,
-  },
   create_mr = {
     target = nil, -- Default branch to target when creating an MR
     template_file = nil, -- Default MR template in .gitlab/merge_request_templates
diff --git a/doc/gitlab.nvim.txt b/doc/gitlab.nvim.txt
index 2d7442f..b37b34f 100644
--- a/doc/gitlab.nvim.txt
+++ b/doc/gitlab.nvim.txt
@@ -230,10 +230,6 @@ you call this function with no values the defaults will be used:
         success = "✓",
         failed = "<U+F467>",
       },
-      merge = { -- The default behaviors when merging an MR, see "Merging an MR"
-        squash = false,
-        delete_branch = false,
-      },
       create_mr = {
         target = nil, -- Default branch to target when creating an MR
         template_file = nil, -- Default MR template in .gitlab/merge_request_templates
diff --git a/lua/gitlab/actions/merge.lua b/lua/gitlab/actions/merge.lua
index 324dc1e..b610c8e 100644
--- a/lua/gitlab/actions/merge.lua
+++ b/lua/gitlab/actions/merge.lua
@@ -17,7 +17,7 @@ end

 ---@param opts MergeOpts
 M.merge = function(opts)
-  local merge_body = { squash = state.settings.merge.squash, delete_branch = state.settings.merge.delete_branch }
+  local merge_body = { squash = state.INFO.squash, delete_branch = state.INFO.delete_branch }
   if opts then
     merge_body.squash = opts.squash ~= nil and opts.squash
     merge_body.delete_branch = opts.delete_branch ~= nil and opts.delete_branch
diff --git a/lua/gitlab/state.lua b/lua/gitlab/state.lua
index 61e0684..594dd72 100644
--- a/lua/gitlab/state.lua
+++ b/lua/gitlab/state.lua
@@ -87,10 +87,6 @@ M.settings = {
       return " " .. discussions_content .. " %#Comment#| " .. notes_content .. help
     end,
   },
-  merge = {
-    squash = false,
-    delete_branch = false,
-  },
   create_mr = {
     target = nil,
     template_file = nil,
```

Then, I think, during a `merge` the MR settings on Gitlab will be respected by default, but it will be possible to override them in the function call `require("gitlab").merge({ squash = false, delete_branch = false })`. Of course, I'd also have to adjust documentation accordingly.

lua/gitlab/actions/create_mr.lua
harrisoncramer: Looks good to me! Can you push up your changes?

harrisoncramer:
@jakubbortlik All of your proposals look good to me, if you push up I'll approve. I'll take a look at adding some documentation for the Lua test suite soon

jakubbortlik:
Hi Harrison, I've applied the three changes as agreed (got rid of `state.settings.merge`, moved metadata in Confirmation popup, and used a new utils function for getting non-nil values), but I've also taken the liberty to modify the `doc/gitlab.nvim.txt` file (commit `8e1ac27e`) to make the formatting more consistent, to make it easier to format the description of API function parameters with the `expandtab` and `sw=4` options. And I've also used "column heading" highlighting with tilde for individual API functions, as previously, I found orientation in the API reference rather difficult. I hope you don't mind these changes.

harrisoncramer (approved):

Issue #177, opened 2024-02-11
Save contents of popup to backup register when quitting


jakubbortlik:
## Feature Description
I'd like to have the contents of a comment popup saved to a backup register when hitting `q` to quit, similar to saving to `backup_register` when `perform_action` is triggered. The purpose would be to temporarily interrupt editing a comment, do some work, and then open the comment again and paste the previously typed contents back in - I'd like to use the unnamed register (`:h quotequote`) so that I could simply paste with `p` without referencing the register. That's why this should be a separate register, other than the current `backup_register` whose purpose it is (in my workflow) to be able to use the finished comment outside of `neovim`, so that the `+` register is more suitable for it.

This would be a kind of primitive "cache" alternative to what is possible in Gitlab online. I'll be glad to open a PR on this. Within the PR, I'd like to make some more improvements:
1. Allow a list of registers to be used (e.g., `backup_register = { "+", "b" }`). This would make the method a little more robust, in case the `+` register got overwritten by some other copy action in/outside of `neovim`. This should not be a breaking change and a string should still be accepted, I guess.
2. If no backup register is defined, save to a global variable instead (something like `vim.g.gitlab_note_backup`), and show information about this to the user, something like "Buffer contents saved to g:gitlab_note_backup. Use :let @+=g:gitlab_note_backup to save to system clipboard. See :h gitlab.nvim.backup-register for how to get rid of this message.") This message could be suppressed in the settings even if not using any backup register. This would address the [issue](https://github.com/harrisoncramer/gitlab.nvim/pull/163#issuecomment-1905739639) raised by @briandipalma that a backup register should be used by default in order to prevent data loss.

harrisoncramer:
I like this idea, but would refine it slightly: Quitting any editable popup buffer before performing an action should save the buffer contents to a predefined register. Couple of things though:

1. I do not think we should introduce a breaking change in the old settings, since it's a completely different feature from the old "backup register" concept. That is a temporary hotfix that I intend to remove completely once we've sorted out the bugs with adding single-line discussions (done) and multi-line discussions (in-progress). It's not permanent.
2. You could call this new setting "temp_register" and make it a list, like you describe.
3. I would introduce a hook into the popup abstraction that does this automatically rather than creating a new "q" keybinding. See the documentation here: https://github.com/MunifTanjim/nui.nvim/tree/main/lua/nui/popup

harrisoncramer:
[...]

PR #238, opened 2024-04-04
Feat: Allow saving popup contents to temp_registers when quitting

jakubbortlik:
This MR addresses #177. The user can now specify a list of registers in `settings.popup.temp_registers` to which the contents of editable popups will be saved when exiting the popup.

A couple of notes on things that might be done differently:
1. If no `temp_registers` are defined, nothing happens. Instead, when quitting a popup without a temp_register being set, the user could be notified about the possibility as mentioned in the original feature request #177.
2. The `temp_registers` are filled even when triggering `settings.popup.perform_action` because this also triggers the `exit` function with the callback `lua/gitlab/state.lua:259` that does the popup saving. If this behaviour is undesirable, the `popup_opts` could be extended to make it possible not to save upon `perform_action`. However, I find this to be a nice side effect to the main purpose of `temp_registers`, as it makes it possible to get rid of the `backup_register` hotfix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ require("gitlab").setup({
     pipeline = nil,
     reply = nil,
     squash_message = nil,
-    temp_registers = {},
+    temp_registers = {}, -- List of registers for backing up popup content (see `:h gitlab.nvim.temp-registers`)
   },
   discussion_tree = { -- The discussion tree that holds all comments
     auto_open = true, -- Automatically open when the reviewer is opened
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ require("gitlab").setup({
     pipeline = nil,
     reply = nil,
     squash_message = nil,
+    temp_registers = {},
   },
   discussion_tree = { -- The discussion tree that holds all comments
     auto_open = true, -- Automatically open when the reviewer is opened
--- a/doc/gitlab.nvim.txt
+++ b/doc/gitlab.nvim.txt
@@ -152,7 +152,7 @@ you call this function with no values the defaults will be used:
         pipeline = nil,
         reply = nil,
         squash_message = nil,
-        temp_registers = {},
+        temp_registers = {}, -- List of registers for backing up popup content (see `:h gitlab.nvim.temp-registers`)
       },
       discussion_tree = { -- The discussion tree that holds all comments
         auto_open = true, -- Automatically open when the reviewer is opened
--- a/doc/gitlab.nvim.txt
+++ b/doc/gitlab.nvim.txt
@@ -152,6 +152,7 @@ you call this function with no values the defaults will be used:
         pipeline = nil,
         reply = nil,
         squash_message = nil,
+        temp_registers = {},
       },
       discussion_tree = { -- The discussion tree that holds all comments
         auto_open = true, -- Automatically open when the reviewer is opened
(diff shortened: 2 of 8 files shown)


jakubbortlik:
Hi @harrisoncramer, what about instead of completely removing the `<ESC>` keybinding we'd just replace it with something that is typically used in plugins to close popup windows, like `gq`? I'm using just `q` and I'm happy with it as I rarely need to record to a register while writing a comment.

With the keymap removed, will the saving to `temp_register` even take place? I'd think we should keep the keymap (and change it from `<ESC>` to something else) even if we'd implement saving to temp files.

jakubbortlik:
By the way, @petRUShka, I'm using `<ZZ>` as my `perform_action` keymap which effectively makes it behave like saving the commit message you mentioned: I "save" the buffer and the comment is applied.

harrisoncramer:
Hey both,

Thanks for the feedback. Believe ultimately that @petRUShka is right, and that the plugin shouldn't be creating new abstractions for things like closing buffers or tabs that are already part of Neovim. 

Instead of having a keybinding for this, users should now just close the popups like they would any other buffer. To address your point @jakubbortlik, I've added an autocommand to the `state.lua` file that will trigger whenever a popup unloads. If you want to add custom saving functionality like in this plugin, this would be the place to do it, as this autocommand is called already when the buffer unloads:

```lua
  vim.api.nvim_create_autocmd("BufUnload", {
    buffer = popup.bufnr,
    callback = function()
      local text = u.get_buffer_text(popup.bufnr)
      -- Do something with the text here...
      exit(popup, opts)
    end,
  })
```

jakubbortlik:
> Hey both, thanks for the feedback.
> 
> The plugin shouldn't be creating new abstractions for things like closing buffers or tabs that are already part of Neovim, so I've removed the ability to add one for popups.
> 
> Instead of having a keybinding for this, users should now just close the popups like they would any other buffer. To address your point @jakubbortlik, I've added an autocommand to the `state.lua` file that will trigger whenever a popup unloads. If you want to add custom saving functionality, this would be the place to do it, as this autocommand is called already when the buffer unloads:
> 
> ```lua
>   vim.api.nvim_create_autocmd("BufUnload", {
>     buffer = popup.bufnr,
>     callback = function()
>       local text = u.get_buffer_text(popup.bufnr)
>       -- Do something with the text here...
>       exit(popup, opts)
>     end,
>   })
> ```
> 
> I still support this MR and think we should add it to the plugin!
> 
> To the other points about saving all comments as drafts, or saving in real-time as the user inputs text, I don't anticipate adding this to the plugin at any point soon, this would actually be very complex to implement and I'm not sure how valuable it'll be. Supporting state for parallel workstreams in different comments/buffers is not a can of worms I'm keen to open up.
> 
> @jakubbortlik Please let me know when this is ready for review and I'll take a look but we should try to get #230 merged first.

Hi @harrisoncramer, thanks for your input. As you say, I'm first focusing on finishing #230 and I'll get back to this after that.

jakubbortlik:
Hi @harrisoncramer, I think the feature works as expected now, although I'd like to suggest one small change to the `autocommand`. If we changed `BufUnload` to `BufLeave`, then a user (me :smile:) would be able to create this mapping to quit a popup without saving the contents to the `temp_registers`:
```lua
vim.keymap.set("n", "ZQ", function()
  local reg_backup = gitlab.state.settings.popup.temp_registers
  gitlab.state.settings.popup.temp_registers = {}
  vim.cmd("quit")
  gitlab.state.settings.popup.temp_registers = reg_backup
end)
```
This is inspired by the behaviour of normal `ZQ` in that it quits "without checking for changes". For some reason, this mapping does not work when `BufUnload` is used. In that case, `ZQ` still saves to the registers.

PR #220, opened 2024-03-11
fix(regex): fix regex for urls with ports #142

Zombiefleischer:
fix #142

PR #248, opened 2024-04-08
ssh remote url can have custom port

sunfuze:
Same as https://github.com/harrisoncramer/gitlab.nvim/pull/220, and add a test case.


harrisoncramer:
Thank you! Can you please target the `develop` branch instead of main? And is this enough for you to connect to a custom port or is there additional configuration required?

If there are more steps we probably just want to call this out in the README for users.

sunfuze:
> Thank you! Can you please target the `develop` branch instead of main? And is this enough for you to connect to a custom port or is there additional configuration required?
Done.
> 
> If there are more steps we probably just want to call this out in the README for users.

No additional configuration is needed, I have used the fork version, it works well.

PR #258, opened 2024-04-10
Feat: Copy web url of MR to clipboard

jakubbortlik:
One less reason to go to the Web :)

This also cosmetically refactors code for getting the node url.


harrisoncramer:
Looks good, thanks! Can we just rename the method to `copy_mr_url()` instead?

jakubbortlik:
> Looks good, thanks! Can we just rename the method to `copy_mr_url()` instead?

No problem. I was blindly copying the name from the `state.INFO`, but `copy_mr_url()` is more appropriate.

PR #259, opened 2024-04-11
Update gitlab.nvim.txt

briandipalma:


jakubbortlik:
Thanks for noticing and fixing this.

PR #260, opened 2024-04-12
Fix: Cursor position outside buffer

jakubbortlik:
This MR fixes a small bug that cause the "cursor position outside buffer" error when `old_line = 0` - `gitlab.nvim` tried to jump to line 0 in the OLD buffer, instead of focusing the NEW buffer.


harrisoncramer:
I just merged in a fix for this, the logic was actually a bit off just generally. Let me know if develop fixes this for you.

jakubbortlik:
Thanks, `develop` is fixes the issue. And I confirm that it also fixes the problem with jumping to the wrong buffer.

PR #265, opened 2024-04-14
Feat: Add keymaps and linewise actions to layouts

jakubbortlik:
This PR improves the UI of layout popups.

1. It makes it possible to cycle through fields in popups (currently Summary and Confirmation pipups), the default keybindings are TAB and S-TAB, they also accept counts (not exceptionally useful with a small number of fields, but maybe there'll be a popup with a larger number of fields in the future). The keybindings can be set up in settings.popup.keymaps - maybe the other keymaps (actions) could be moved there as well, after a depreciation period, to keep the settings organized. Similarly, the discussion tree keymaps could be moved to a separate table. Now they are not very clearly distinguished from other tree settings.
2. Linewise actions have been added for the Boolean fields and the target branch field in the confirmation popup for MR creation.
3. Keymaps for the info field in the summary are added in order to show help with `g?`. I'd like to add a linewise action for changing the target branch here also, but this will require changes to the go server. I'm not sure if go-gitlab makes it possible to update the squash and delete_branch fields, though.


lua/gitlab/actions/miscellaneous.lua
```diff
@@ -38,4 +38,56 @@ M.editable_popup_opts = {
...
+---@param popups table Table of Popups
+M.set_cycle_popups_keymaps = function(popups)
+  local number_of_popups = #popups
+  for i, popup in ipairs(popups) do
+    popup:map("n", state.settings.popup.keymaps.next_field, function()
+      vim.api.nvim_set_current_win(popups[next_index(i, number_of_popups, vim.v.count)].winid)
+    end, { desc = "Go to next field (accepts count)" })
+    popup:map("n", state.settings.popup.keymaps.prev_field, function()
+      vim.api.nvim_set_current_win(popups[prev_index(i, number_of_popups, vim.v.count)].winid)
+    end, { desc = "Go to previous box (accepts count)" })
```
jakubbortlik, line 80:
```suggestion
    end, { desc = "Go to previous field (accepts count)" })
```

lua/gitlab/utils/init.lua
```diff
@@ -656,6 +695,27 @@ M.get_all_git_branches = function(remote)
   return branches
 end
 
+---Select a git branch, add it to callback args and perform callback
+---@param cb function The call back to perform with the selected branch
+---@param args table Table with callback arguments
+---@param arg string Name under which the selected branch will be added to the args
+---@param remote? boolean Whether remote-tracking branches should be listed
+---@param exclude_current? boolean Whether the current branch should be excluded from the list
+M.select_target_branch = function(cb, args, arg, remote, exclude_current)
```
harrisoncramer, line 704: We are only supporting opening up MRs against branches that are on the remote since that's all that Gitlab can do, this function doesn't need any arguments. Please also look into using the `List` module in the future for logic like this, it makes things easier to understand. I'm trying to get away from so many C-style for loops which are difficult to follow.

harrisoncramer: I think this is all that's needed here, no need for all the `args` and booleans:

```lua
M.get_all_merge_targets = function()
  local handle = io.popen("git branch -r 2>&1")
  if not handle then
    M.notify("Error running 'git branch' command.", vim.log.levels.ERROR)
    return
  end

  local current_branch = M.get_current_branch()
  if not current_branch then
    return
  end

  local lines = {}
  for line in handle:lines() do
    table.insert(lines, line)
  end
  handle:close()

  -- Trim "origin/" and don't include the HEAD pointer
  local branches = List.new(lines)
      :map(function(line)
        return line:match("origin/(%S+)")
      end)
      :filter(function(branch)
        return not branch:match("^HEAD$") and branch ~= current_branch
      end)

  return branches
end

---Select a git branch
---@param cb function The call back to perform with the selected branch
M.select_target_branch = function(cb)
  local all_branch_names = M.get_all_merge_targets()
  if not all_branch_names then
    return
  end
  vim.ui.select(all_branch_names, {
    prompt = "Choose target branch for merge",
  }, function(choice)
    if choice then
      cb(choice)
    end
  end)
end
```

harrisoncramer (changes requested):
This works very well, just requested some changes on the changes related to MR selection.

lua/gitlab/utils/init.lua
jakubbortlik: Hi Harrison, thanks for your suggestions. It didn't occur to me to use the `List` functionality but it will certainly make things clearer. I wanted to keep the `remote` parameter and introduced the `exclude_current` parameter in `get_all_git_branches` to really make it possible to get *all* git branches. But as this is not really necessary for the plugin, I'll remove those two parameters.

I'm not sure, however, about the `args` parameter. The reason I used it is that the `utils.select_target_branch` method replaces this piece of code:
```lua
  local all_branch_names = u.get_all_git_branches(true, true)
  vim.ui.select(all_branch_names, {
    prompt = "Choose target branch for merge",
  }, function(choice)
    if choice then
      mr.target = choice
      M.pick_template(mr)
    end
  end)
```
As you can see, the `choice` of `vim.ui.select` was not actually passed to the `callback` method (`create_mr.pick_template`) directly, but it was only added to the other arguments, i.e., `mr = {target = "THE PICKED TARGET", delete_branch = true, ...}`. I'll try to figure something out.

jakubbortlik: I've made the `Delete branch`, `Squash commits` and `Target branch` fields non-modifiable to prevent users from inserting invalid values. I was also thinking about adding an autocommand that would show help if a user attempted to modify the buffers, something like:
```lua
  vim.api.nvim_create_autocmd("CursorHold", {
    buffer = buf,
    callback = function()
      if vim.v.errmsg == "E21: Cannot make changes, 'modifiable' is off" then
        M.notify("Do not edit manually. Use " .. require("gitlab.state").settings.popup.perform_linewise_action .. " instead.", vim.log.levels.INFO)
        vim.v.errmsg = ""
      end
    end,
  })
```
But that's probably superfluous.
Now I'm rather considering whether to simplify the `toggle_string_bool` method in `lua/gitlab/utils/init.lua:377`, since it does not need all the case and whitespace checking anymore.

harrisoncramer: Looks like you requested another review but I don't think you pushed up your changes

harrisoncramer (approved):

harrisoncramer:
Thank you @jakubbortlik as always! This plugin wouldn't be what it is without your contributions.

jakubbortlik:
> Thank you @jakubbortlik as always! This plugin wouldn't be what it is without your contributions.

I really like contributing to the amazing tool you've built! I use the plugin daily at work and it just makes my life so much easier. So I'm glad I can help improve it.

Issue #249, opened 2024-04-08
Pipeline API is (inconsistently) Failing


harrisoncramer:
## Prerequsities

- [x] I'm on the latest version of the plugin
- [x] I've installed the required dependencies
- [x] I've run `:h gitlab.nvim.troubleshooting` and followed the steps there

## Setup Configuration and Environment

For some reason the call to get the latest pipeline is failing sometimes for me, but it's not consistent. It varies across different branches. I've asked about it in Gitlab's Reddit community here: https://www.reddit.com/r/gitlab/comments/1bywlze/inconsistent_permission_errors_on_gitlabs/

## Bug Description

There are 403 errors when getting the latest pipeline for some branches. This may be an issue with how my tenant is configured, not sure.

## Reproduction Steps

1. Check out the feature branch
2. Open Neovim
3. Run the `require("gitlab").pipeline()` command.

harrisoncramer:
Addresses #249, for whatever reason the API to get the latest pipeline for a given MR is not working for me, I'm getting 403 errors. However, using the project-wide pipelines endpoint to get the latest pipeline (and filtering for the most recent commit on the current branch) _does_ work. This MR changes the pipeline API call to that strategy.

PR #266, opened 2024-04-14
fix: Get recent pipeline through other means

harrisoncramer:


harrisoncramer:
Addresses https://github.com/harrisoncramer/gitlab.nvim/issues/249, for whatever reason the API to get the latest pipeline for a given MR is not working for me, I'm getting 403 errors. However, using the project-wide pipelines endpoint to get the latest pipeline (and filtering for the most recent commit on the current branch) does work. This MR changes the pipeline API call to that strategy.

Issue #263, opened 2024-04-12
Diagnostic error due to missing line numbers in discussion data with `position_type` "file"


jakubbortlik:
## Prerequsities

- [x] I'm on the latest version of the plugin
- [x] I've installed the required dependencies
- [x] I've run `:h gitlab.nvim.troubleshooting` and followed the steps there

## Bug Description

In the function `create_single_line_diagnostic` (`lua/gitlab/indicators/diagnostics.lua:49`), it is assumed, that at least one of new line or old line is set. If both are `nil`, the following error is thrown: `gitlab.nvim: Error setting diagnostics: ...ojects/gitlab.nvim/lua/gitlab/indicators/diagnostics.lua:52: attempt to perform arithmetic on field 'old_line' (a nil value)` (along with around 200 lines of other errors from `gitlab.nvim` and `diffview` due to some async callbacks and autocommands).

This is the problematic line (line 52 as of `7d76187cdbef54ddfefda7c3b348e6c61744c9eb`):
```lua
    lnum = (first_note.position.new_line or first_note.position.old_line) - 1,
```

## Additional information
This is the relevant part from the log, as you can see, the lines numbers are missing completely:
```json
{
    "position": {
        "base_sha": "324c9ff79aa7cf5ca7006f4641325cf044e616d4",
        "start_sha": "324c9ff79aa7cf5ca7006f4641325cf044e616d4",
        "head_sha": "a815ff6f4a666fb2a13848b833f86b6c8e062811",
        "old_path": "...",
        "new_path": "...",
        "position_type": "file"
    }
}
```
Notice also the `"position_type": "file", here's a [gitlab issue](https://gitlab.com/gitlab-org/gitlab/-/issues/423046) regarding this type. And here's a screenshot from Gitlab that shows where you can create such a comment, if you wanted to replicate it:
![image](https://github.com/harrisoncramer/gitlab.nvim/assets/17056143/954a0436-df27-4571-a7f2-bbd09db5791d)

PR #268, opened 2024-04-16
Fix: File comments don't have line number information

jakubbortlik:
Addresses #263. This is a simple fix, not sure whether the missing line number information for "file" comments should be handled in another place.


harrisoncramer:
Looks good, will merge this and ship it with the next release. Who knows why Gitlab has so many different types of comments 🙄

PR #270, opened 2024-04-18
Add authentication provider setting

### LINKED ISSUE #269: Provide setting option for auth provider
## Feature Description
It would be nice if instead of having my gitlab token in either the env variable or properties file, I can specify a lua callback
to implement getting the password myself.

Something along the line of 

```lua
    require("gitlab").setup({
      auth_provider = function()
        return "my_token", "https://custom.gitlab.instance.url"
      end,
    }
```
(Note: I have a PR ready, so if this is something you do not want in, feel free to decline it 😅 )

patrickpichler:
It is now possible to configure an authentication provider, that will be used to retreive the gitlab token via a lua function.

Fixes #269

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,18 @@ gitlab_url=https://my-personal-gitlab-instance.com/
 
 The plugin will look for the `.gitlab.nvim` file in the root of the current project by default. However, you may provide a custom path to the configuration file via the `config_path` option. This must be an absolute path to the directory that holds your `.gitlab.nvim` file.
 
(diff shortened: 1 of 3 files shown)


harrisoncramer:
Looks good, thanks for doing this. Can you please target the `develop` branch? We use that as a staging area before merging into main to ensure that we aren't breaking any existing functionality.

Please also run `stylua` to fix formatting.

patrickpichler:
Target branch has been updated and `stylua` has been run.

PR #281, opened 2024-04-22
Feat: Make toggle_draft_mode available outside discussion tree

jakubbortlik:
It's cool that the draft mode can be toggled without restarting Neovim. This PR makes this functionality available for mapping outside of the discussion tree.

Also, this PR some `gitlab.toggle_resolved`, and `gitlab.reply` as global actions, since they require a `tree` as a parameter, and they do not work when run like `require("gitlab").reply()`.

### DIFF
--- a/lua/gitlab/init.lua
+++ b/lua/gitlab/init.lua
@@ -74,7 +74,7 @@ return {
     draft_notes_dep,
     discussion_data,
   }, discussions.toggle),
-  toggle_draft_mode = async.sequence({ info }, discussions.toggle_draft_mode),
+  toggle_draft_mode = discussions.toggle_draft_mode,
   publish_all_drafts = draft_notes.publish_all_drafts,
   -- Other functions 🤷
   state = state,
--- a/lua/gitlab/init.lua
+++ b/lua/gitlab/init.lua
@@ -74,9 +74,8 @@ return {
     draft_notes_dep,
     discussion_data,
   }, discussions.toggle),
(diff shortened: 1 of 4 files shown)


lua/gitlab/init.lua
```diff
@@ -74,9 +74,8 @@ return {
     draft_notes_dep,
     discussion_data,
   }, discussions.toggle),
-  toggle_resolved = async.sequence({ info }, discussions.toggle_discussion_resolved),
+  toggle_draft_mode = async.sequence({ info }, discussions.toggle_draft_mode),
   publish_all_drafts = draft_notes.publish_all_drafts,
-  reply = async.sequence({ info }, discussions.reply),
```
harrisoncramer, line 79: Don't think this is quite right, you don't want to be getting rid of these other mappings. You also don't need to wrap this in `async.sequence` which is used to fetch data via the Gitlab API.

jakubbortlik: I've removed the unnecessary async call. But I still don't see how `discussions.toggle_discussion_resolved` and `discussions.reply`, or `gitlab.toggle_resolved` and `gitlab.reply`, respectively, could be used outside of the plugin. These functions are not even described in the API documentation. How would a user supply the necessary `tree` argument? So I thought they were included in `lua/gitlab/init.lua` by mistake. If you want, I can get them back, of course.

harrisoncramer: oh yeah you're right, good call

Issue #267, opened 2024-04-15
Add command(s) to refresh discussions


jakubbortlik:
## Feature Description

I'd like to have a command (and ideally also a keybinding) for refreshing the discussion tree. Currently, if I want to refresh the tree, I have to toggle it twice, which is not practical - first of all, why run a command twice, if it could be just one command. But also, sometimes I modify the layout (make the vertically split discussion tree wider, etc.) and this change is lost when toggling. I'd just like to refresh the data.

I think the default keybinding for this inside the discussion tree could be `CTRL-R` as it is otherwise unused in the non-modifiable buffer. Alternatively, there could be a command that would work from anywhere in `nvim` and that users could map to whatever they liked, just like the other commands, e.g., `vim.keymap.set("n", "gl<c-r>", gitlab.refresh_discussions)`.

I can create a PR for this.

jakubbortlik:
What I'd actually like to achieve is that the **nodes would not collapse** when refreshing the discussion tree and when adding or editing nodes. I've tried for some time to get this behaviour, but cannot figure it out. Interestingly, for deleting and adding replies, the tree behaves just the way I like - the toggling status of the tree is not changed. But when editing/adding nodes, the tree is completely rebuilt. I assume this has to do with the Go library and the Gitlab API, right? I'm afraid I wouldn't easily find a way how to get this done.

harrisoncramer:
This will be addressed when #288 ships 🥳

PR #288, opened 2024-04-23
feat: allow refreshing discussion data

harrisoncramer:
Address #267.

Introduces a uniform way to refresh the discussions view via a uniform The GET-based refresh API. The discussion tree is no longer mutated on a case-by-case basis by edit, reply, and other endpoints. Instead, after those actions are completed, the appropriate resources are re-fetched from the server. Although this makes the API _slightly_ less responsive, the application is more stable as a result and the state is guaranteed to match the state in Gitlab at all times.

This MR therefore also addresses the issue of the tree collapsing following a change. The default keybinding to refresh data in the discussion tree will be `a` moving forward.


lua/gitlab/actions/draft_notes/init.lua
```diff
@@ -92,67 +98,18 @@ M.send_edits = function(note_id)
...
-    else
-      discussions.rebuild_unlinked_discussion_tree()
-    end
-
-    if state.settings.discussion_signs.enabled and state.DISCUSSION_DATA then
-      diagnostics.refresh_diagnostics()
-    end
-
-    winbar.update_winbar()
-    common.add_empty_titles()
```
harrisoncramer, line 155: Notice all of this code is deleted 🥳 

It's now no longer necessary to parse the returned state, we just dumbly re-fetch data from our GET endpoints and rebuild the tree from scratch. This accounts for a lot of deleted code in this MR.

PR #297, opened 2024-05-03
fix(mr.create): consider remote branches

### LINKED ISSUE #296: Create MR branch logic does not include remote branches
## Prerequsities

## Bug Description

`get_all_merge_targets` -> `get_all_remote_branches` -> `branches` calls `git branch`, which is seemingly incorrect if we are trying to fetch remote branches - should use `git branch --remotes` or `git branch --all`.

I have thrown in some debug printing strings to enumerate the issue, as follows:

```
# Listing the current branch
[user] -- DBGSTR [git.lua:52-7] current_branch=test-branchy

# Output from `git branch` as called by M.branches()
[user] -- DBGSTR [git.lua:69-9] all_branches=master
[user] * test-branchy

# Parsing of string to Lua table
[user] -- DBGSTR [git.lua:79-10] lines={ "master", "* test-branchy" }
```

Then the following code:

```lua
  return List.new(lines)
    :map(function(line)
      -- Trim "origin/"
      return line:match("origin/(%S+)")
    end)
    :filter(function(branch)
      -- Don't include the HEAD pointer
      return not branch:match("^HEAD$")
    end)
```

is obviously going to skip all of these lines, as none of them match the patterns (the first filter filters out all branches, and the second pattern has nothing to match against).

If I change the call to be `git branch --remotes`, it looks more like this:

```
# Listing the current branch
[user] -- DBGSTR [git.lua:54-8] remotes={ { "master", "test-branchy" } }

# Output from `git branch` as called by M.branches() (with --remotes)
[user] -- DBGSTR [git.lua:69-9] all_branches=master
[user] * test-branchy
[user]   remotes/origin/master
[user]   remotes/origin/test-branchy

# Parsing of string to Lua table
[user] -- DBGSTR [git.lua:79-10] lines={ "master", "* test-branchy", "  remotes/origin/master", "  remotes/origin/test-branchy" }

# Comparisons later on which show that some branches actually get returned (above no branches were returned)
[user] -- DBGSTR [git.lua:57-5] branch=master
[user] -- DBGSTR [git.lua:58-6] current_branch=test-branchy
[user] -- DBGSTR [git.lua:57-5] branch=test-branchy
[user] -- DBGSTR [git.lua:58-6] current_branch=test-branchy
```

levouh:
Fix #296

### DIFF
--- a/lua/gitlab/git.lua
+++ b/lua/gitlab/git.lua
@@ -1,4 +1,5 @@
 local List = require("gitlab.utils.list")
+local u = require("gitlab.utils")
 
 local M = {}
 
@@ -6,7 +7,6 @@ local M = {}
 ---@param command table
 ---@return string|nil, string|nil
 local run_system = function(command)
-  local u = require("gitlab.utils")
   local result = vim.fn.trim(vim.fn.system(command))
   if vim.v.shell_error ~= 0 then
     u.notify(result, vim.log.levels.ERROR)
@@ -16,9 +16,10 @@ local run_system = function(command)
 end
 
 ---Returns all branches for the current repository
+---@param args table|nil extra arguments for `git branch`
 ---@return string|nil, string|nil
-M.branches = function()
-  return run_system({ "git", "branch" })
+M.branches = function(args)
+  return run_system(u.combine({ "git", "branch" }, args or {}))
 end
 
 ---Checks whether the tree has any changes that haven't been pushed to the remote
@@ -60,7 +61,7 @@ end
 ---Return the list of names of all remote-tracking branches or an empty list.
 ---@return table, string|nil
 M.get_all_remote_branches = function()
-  local all_branches, err = M.branches()
+  local all_branches, err = M.branches({ "--remotes" })
   if err ~= nil then
     return {}, err
   end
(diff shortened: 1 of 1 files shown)


lua/gitlab/git.lua
```diff
@@ -16,9 +17,12 @@ local run_system = function(command)
 end
 
 ---Returns all branches for the current repository
+---@param args table|nil extra arguments for `git branch`
 ---@return string|nil, string|nil
-M.branches = function()
-  return run_system({ "git", "branch" })
+M.branches = function(args)
+  -- Load here to prevent loop
+  local u = require("gitlab.utils")
```
levouh, line 24: Not sure if it is worth adding a `require_on_index` implementation to `gitlab.nvim`, as I am not sure how often this is actually an issue (require loops). But this can also be solved by adding something like:

```lua
M.require_on_index = function(require_path)
  return setmetatable({}, {
    __index = function(_, key)
      return require(require_path)[key]
    end,

    __newindex = function(_, key, value)
      require(require_path)[key] = value
    end,
  })
end
```

and then doing:

```lua
local u = require("gitlab.util.lazy").require_on_index("gitlab.utils")
```

And since it is not required until it is used, you do not run into the same require looping stuff that causes this to be necessary.

Issue #302, opened 2024-05-27
Bug: Ranged Comment Signs Not Showing (Neovim .10)


harrisoncramer:
## Prerequsities

- [x] I'm on the latest version of the plugin
- [x] I've installed the required dependencies
- [x] I've run `:h gitlab.nvim.troubleshooting` and followed the steps there

## Bug Description

When creating a multi-line diagnostic in Neovim `.10` only the sign related to the discussion is placed. The ranged signs are not placed.

This is happening inside of the `set_signs` function in `lua/gitlab/indicators/signs.lua`

I do not see anything about regressions in `.10` related to diagnostics so not sure where this is coming from...

## Reproduction Steps

1. Check out the feature branch
2. Open Neovim (`.10`)
3. Place a ranged comment

PR #305, opened 2024-06-01
Fix: Ranged comment signs not showing

jakubbortlik:
Addresses #302.

This PR also makes it possible to hide the diagnostic sign (e.g., `I` if the diagnostic level is "INFO") - this makes the diagnostic range icons better aligned. The default behaviour of the plugin does not change, i.e., the diagnostic signs are shown.

I've noticed one small bug that existed already with nvim 0.9, if some signs overlap, the `GitlabComment` icons are hidden by the `GitlabRange` icons the first time the signs are placed. This is because for some reason, this line in `lua/gitlab/indicators/signs.lua:33` does not return any already placed signs:
```lua
    local existing_signs =
      vim.fn.sign_getplaced(vim.api.nvim_get_current_buf(), { group = discussion_sign_name })[1].signs
```
On subsequent runs (e.g., when jumping to a diagnostic or when switching files in the review), the overlapping signs are placed correctly and the `GitlabComment` sign takes precedence. If I found the reason for this behaviour I would like to fix it in this PR, but I haven't been successful so far. Maybe it has to do with calling a `vim` function inside a `lua` script.

Regarding the small modifications of the documentation, there are still some inconsistencies between the default setup settings and the ones described in the README and docs (e.g., `skip_old_revision_discussion`), but right now I don't have the time to fix those. Have you considered keeping the settings in let's say just two places? The README could maybe only contain a link to the docs. It would be easier to maintain consistency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,7 @@ require("gitlab").setup({
     skip_resolved_discussion = false, -- Show diagnostics for resolved discussions
     severity = vim.diagnostic.severity.INFO, -- ERROR, WARN, INFO, or HINT
     virtual_text = false, -- Whether to show the comment text inline as floating virtual text
+    use_diagnostic_signs = true, -- Show diagnostic sign (depending on the `severity` setting, e.g., I for INFO) along with the comment icon
     priority = 100, -- Higher will override LSP warnings, etc
     icons = {
       comment = "→|",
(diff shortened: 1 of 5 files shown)


jakubbortlik:
Maybe the issue with incorrect signs could be solved by using the `nvim_buf_set_extmark` function instead of the `sign_place` function that is called "legacy" in the news for nvim 0.10.

harrisoncramer:
The remark about linking to the settings block is a great idea. I'll add an MR to do that.

Issue #308, opened 2024-06-03
Get error 'init.lua:298: Failed to unload buffer' when post summary description for MR


sergey-arkhipov:
## Prerequsities

- [x] I'm on the latest version of the plugin
- [x] I've installed the required dependencies
- [x] I've run `:h gitlab.nvim.troubleshooting` and followed the steps there

## Setup Configuration and Environment

Default configuration

## Bug Description

When post summary for MR get error

```bash
Error detected while processing BufWinLeave Autocommands for "<buffer=5>":
E937: Attempt to delete a buffer that is in use: [No Name]
Error executing lua callback: ...s/.local/share/nvim/lazy/nui.nvim/lua/nui/popup/init.lua:298: Failed to unload buffer.
stack traceback:
  [C]: in function 'nvim_buf_delete'
  ...s/.local/share/nvim/lazy/nui.nvim/lua/nui/popup/init.lua:298: in function '_buf_destroy'
  ...s/.local/share/nvim/lazy/nui.nvim/lua/nui/popup/init.lua:322: in function 'unmount'
  .../.local/share/nvim/lazy/gitlab.nvim/lua/gitlab/state.lua:270: in function 'exit'
  .../.local/share/nvim/lazy/gitlab.nvim/lua/gitlab/state.lua:319: in function <.../.local/share/nvim/lazy/gitlab.nvim/lua/gitlab/state.lua:312>
  [C]: in function 'nvim_buf_delete'
  .../.local/share/nvim/lazy/gitlab.nvim/lua/gitlab/state.lua:293: in function <.../.local/share/nvim/lazy/gitlab.nvim/lua/gitlab/state.lua:289>

```

## Reproduction Steps

1. Check out the feature branch
2. Open Neovim
3. Add some description for MR
4. POST (save)

## Screenshots

![Screenshot 2024-06-03 at 17 42 33](https://github.com/harrisoncramer/gitlab.nvim/assets/5460458/fcd6d84e-8124-42b4-8d6a-a6c5ef14f03d)

harrisoncramer:
Just to be clear, is this when adding a description to an existing MR, or is this creating a new MR? 

And when you say `POST (save)` what specific action are you taking -- are you doing a primary action with the MR description open, in order to save and send to Gitlab? I'm not able to replicate this so any more detail you can provide will be helpful.

sergey-arkhipov:
> any more detail you can provide will be helpful

With peasure,

Simple steps.

[...]

PR #315, opened 2024-06-19
Fix: Failure when unloading buffer

jakubbortlik:
This PR addresses #308.

The code for closing popups is quite complicated, so it would still benefit from a refactor. The `popup:unmount()` methods are called necessarily several times. This PR gets rid at least of some of these repeated calls and it also gets rid of the call to `vim.api.nvim_buf_delete()` in favour of the existing `exit(popup, opts)` function which has the advantage that the `nui` library makes sure there is something to be unmounted when `layout:unmount` or `popup:unmount` are called.

This PR also limits the `BufWinLeave` autocommands to just the buffers that actually need them.

### DIFF
--- a/lua/gitlab/actions/summary.lua
+++ b/lua/gitlab/actions/summary.lua
@@ -63,10 +63,20 @@ M.summary = function()
       description_popup,
       M.edit_summary,
       miscellaneous.attach_file,
-      { cb = exit, action_before_close = true, save_to_temp_register = true }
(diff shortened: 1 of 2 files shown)


harrisoncramer:
This doesn't _break_ anything for me but maybe let's see if this resolves the commenter's issue before we merge it in. I've not had any issues regarding closing popups or anything. Thanks for working on this.

PR #316, opened 2024-06-19
Fix docs: choose_merge_request's open_reviewer default value is true

eclbg:
Hi Harry,
Thanks so much for writing this plugin. I've recently started a job at a company that uses Gitlab so I'm giving it a go. I noticed that the default value of `open_reviewer` when calling `choose_merge_request` was wrong in the docs, so here's a tiny PR to fix it.
Cheers!

### DIFF
--- a/doc/gitlab.nvim.txt
+++ b/doc/gitlab.nvim.txt
@@ -635,7 +635,7 @@ This command will automatically check out that branch locally, and optionally
 open the reviewer pane. This is the default behavior.
 >lua
   require("gitlab").choose_merge_request()
-  require("gitlab").choose_merge_request({ open_reviewer = false })
+  require("gitlab").choose_merge_request({ open_reviewer = true })
 <
     Parameters: ~
         • {opts}: (table|nil) Keyword arguments to configure the checkout.


harrisoncramer:
Thanks!

jakubbortlik:
Hi @harrisoncramer, I don't think the `require("gitlab").choose_merge_request({ open_reviewer = false })` was a mistake -- it was supposed to demonstrate how to _override_ the default behaviour. @eclbg probably understood "This is the default behaviour" as saying: "the following code shows you the default behaviour". I think the following would be a more appropriate change:
```diff
diff --git a/doc/gitlab.nvim.txt b/doc/gitlab.nvim.txt
index c7af27f..426a866 100644
--- a/doc/gitlab.nvim.txt
+++ b/doc/gitlab.nvim.txt
@@ -631,11 +631,12 @@ default arguments outlined under "Configuring the Plugin".
 gitlab.choose_merge_request({opts}) ~
 
 Choose a merge request from a list of those open in your current project to review. 
-This command will automatically check out that branch locally, and optionally
-open the reviewer pane. This is the default behavior.
+This command will automatically check out that branch locally, and by default
+also open the reviewer pane (this can be overridden with the `open_reviewer`
+parameter).
 >lua
   require("gitlab").choose_merge_request()
-  require("gitlab").choose_merge_request({ open_reviewer = true })
+  require("gitlab").choose_merge_request({ open_reviewer = false })
 <
     Parameters: ~
         • {opts}: (table|nil) Keyword arguments to configure the checkout.
```

eclbg:
Thanks for merging it! 
@jakubbortlik I see your point, and agree that it would be a valid solution too. I have no attachment to my proposal so feel free to go with whatever you prefer @harrisoncramer.
Thanks again

PR #331, opened 2024-07-15
Feat: Add default keymaps

### LINKED ISSUE #287: Set default reviewer keymaps
## Feature Description
I'd like to share a piece of my `gitlab.nvim`/`DiffView` configuration. It creates operator and visual mode mappings in the reviewer windows. Rather than having this in my configuration, I think it would be more suitable to be part of `gitlab.nvim` itself for the following reasons:
1. Comment/suggestion creation and `move_to_discussion_tree_from_diagnostic` only make sense in the reviewer. No need for the mappings to be global.
2. My current configuration creates the mappings for any `DiffView` reviewer buffers, even those not created by `gitlab.nvim`, where they make no sense.
3. Currently, the "plugin does not set up any keybindings outside of the special buffers it creates", but most users probably do create mappings for commenting. However, some of the suggested mappings are quite cumbersome, e.g., to create a suggestion for a paragraph, one would type `vipgls`. With the operator mappings, I can just type `sip`, or `vips` if I want to first see the visual selection -- since the reviewer windows are non-modifiable, `s` and `c` are practically free to remap and the `cip`, `sip` mappings are much easier to type.
4. Using the same keymap for `move_to_discussion_tree_from_diagnostic` as for `jump_to_reviewer` it's easier to quickly move between the tree and reviewer without having to think if its `m` or `glm` that one has to use.
```lua
    ---@param cb string Name of the API function to call
    local function execute_operatorfunc(cb)
      local old_opfunc = vim.opt.operatorfunc
      local cur_win = vim.api.nvim_get_current_win()
      local cur_pos = vim.api.nvim_win_get_cursor(cur_win)

      _G.CreateOperatorfunc = function(callback)
        return function()
          vim.cmd.execute([["normal! '[V']"]])  -- Make sure we are in linewise visual mode
          vim.api.nvim_command(('lockmarks lua require("gitlab").%s()'):format(callback))
          vim.api.nvim_win_set_cursor(cur_win, cur_pos)
          vim.opt.operatorfunc = old_opfunc
        end
      end

      vim.opt.operatorfunc = ("v:lua.CreateOperatorfunc'%s'"):format(cb)
      vim.api.nvim_feedkeys("g@", "n", false)
    end

    require("diffview").setup({
      keymaps = {
        view = {
          { "n", "c", function() execute_operatorfunc("create_multiline_comment") end, { desc = "Create comment in range of motion"} },
          { "n", "s", function() execute_operatorfunc("create_comment_suggestion") end, { desc = "Create suggestion for range of motion"} },
          { "n", "a", function() require("gitlab").move_to_discussion_tree_from_diagnostic() end, { desc = "Move to discussion"} },
          { "v", "s", function () require("gitlab").create_comment_suggestion() end, {desc = "Create suggestion for selected text"}},
          { "v", "c", function () require("gitlab").create_multiline_comment() end, {desc = "Create comment for selected text"}},
        }
      },
    })
```
I'd prefer to use `a` rather than `m` for jumping between the reviewer and the tree since `a` is "free" in the non-modifiable buffers, and `m` could be left available for setting marks. `x` would also be a fine alternative. But of course, I can do this in my personal config, and I can use the existing default `m` for jumping. The default operator mappings could be `glc` and `gls` instead of `c` and `s`, to be more in line with the current mappings, but I find that unnecessarily conservative.

@harrisoncramer, would you be willing to include something like this into the plugin? I imagine this could be called in the `initialize_discussions` function, something like:
```diff
diff --git a/lua/gitlab/actions/discussions/init.lua b/lua/gitlab/actions/discussions/init.lua
index 3894bf5..739ef08 100644
--- a/lua/gitlab/actions/discussions/init.lua
+++ b/lua/gitlab/actions/discussions/init.lua
@@ -55,6 +55,7 @@ M.initialize_discussions = function()
   end)
   reviewer.set_callback_for_reviewer_enter(function()
     M.modifiable(false)
+    M.set_reviewer_keymaps()
   end)
   reviewer.set_callback_for_reviewer_leave(function()
     signs.clear_signs()
```
Where `M.set_reviewer_keymaps` would work similar to `M.modifiable` (`lua/gitlab/actions/discussions/init.lua:68`).

I'd only use `create_multiline_comment` and `create_comment_suggestion`, since `create_comment` is just a special case of `create_multiline_comment` anyway, and it can be achieved by simply using a motion that moves on the same line, e.g., `cl`.

jakubbortlik:
Resolves #287.

In this PR I introduce the following changes:
- keymap settings are all grouped together in a common field in the config table
- keymaps can be disabled globally or individually
- global keymaps for most gitlab.nvim API functions are set
- keymaps for common actions in the reviewer windows are set
- warnings are shown for removed keymap settings fields

Some of the default keybindings in this PR are different from the current defaults. The following are the differences and the reasons for the suggested changes. Some of them are more useful than others and I can revert any of those to the current default in `main` if you prefer to keep the breaking changes to the minimum. Of course I can remap any of these in my personal config. However, I think that some of these suggested mappings nicely override the behaviour of builtin normal mode commands in non-modifiable buffers and that other users might like these mappings too. It would be nice if more users could review this PR and express their opinions about the suggested changes.

### Global keybindings
- `choose_merge_request`: mapping changes from `glb` to `glc` as this is a better mnemonic and is now free for mapping since "glc" is no longer needed for a global `create_comment` mapping
- the funciton `gitlab.review` has the `start_review` keymap field and the mapping changes from `glr` to `glS`. `glr` is the common prefix for `add_reviewer` and `delete_reviewer`, so with `glr` alone the user needs the wait for `timeoutlen` for `gitlab.review` to be called. `start_review` and `glS` are more explicit.
- `create_mr`: is mapped to `glC` as this is a clearer mnemonic than `glO`

### Popup
- `perform_action` is mapped to `ZZ` as the bulitin behaviour of `ZZ` is not needed (we don't really need to _write_ the current file for the popup windows), but I think that it nicely captures what we want to do: apply changes and close the window. This is similar how the `commit` buffers are handled when vim is the default editor for git. ~I'm also planning to add the `ZQ` mapping that would close the window without saving the buffer contents to the `temp_registers`, so that it would logically complement `ZZ`.~ I've also added `ZQ` to close the popup without saving to `temp_registers` so that it nicely complements `ZZ`.

### Discussion tree
- `jump_to_reviewer` changes from `m` to `a` - The point is to have the same keymap for jumping back and forth between the reviewer and discussion tree, but I'd like to keep the `m` for setting marks in the reviewer.
- `refresh_data` changes from `a` to `<c-r>` - this is a common keybinding for refreshing in browsers and other applications. It's a modifying keybinding, so it's free to remap in a non-modifiable buffer.
- `switch_view` changes from `S` to `c` - I'd like to add some functionality around accepting suggestions for which the `S` will be a more suitable mnemonic. We could change `switch_view` to `change_view` to make this one also more natural.
- `toggle_resolved` changes from `p` to `-`. I think it might be useful to add some _preview_ functionality to the discussion tree, which would make it possible to for example see suggestion diffs. For this the `p` would be a nice mnemonic. `-` is used for toggling in other plugins, e.g., fugitive.

## Other notes
- Some of the code that I've added is quite repetitive - adding global keymaps could probably be refactored to use a loop and get rid of some of the duplication, but I guess this can be done in a separate issue.
- For the reviewer keymaps, I've only used the functions `create_comment_suggestion` and `create_multiline_comment`, since `create_comment` for one line is just a special case of a "multiline" comment. I think in the future we could remove the `create_comment` function altogether to reduce the code complexity. What do you think?


jakubbortlik:
One problem with the newly set reviewer keymaps is that they are **buffer**-local but not **window**-local ([window-local mappings](https://github.com/neovim/neovim/issues/16263) are not possible in (n)vim). That means if I'm working on a MR and want to modify one of the files open in the reviewer, I press DiffView's mapping `gf`, I get the file opened in a separate tab, but the reviewer mappings are still in place, which is not desirable - I really want `create_comment/suggestion` to be mapped to `s` and `c`, but then I cannot use these keys to modify the file.

I'll try to fix this with the help of autocommands.

jakubbortlik:
> One problem with the newly set reviewer keymaps is that they are **buffer**-local but not **window**-local ([window-local mappings](https://github.com/neovim/neovim/issues/16263) are not possible in (n)vim). That means if I'm working on a MR and want to modify one of the files open in the reviewer, I press DiffView's mapping `gf`, I get the file opened in a separate tab, but the reviewer mappings are still in place, which is not desirable - I really want `create_comment/suggestion` to be mapped to `s` and `c`, but then I cannot use these keys to modify the file.
> 
> I'll try to fix this with the help of autocommands that would fire when entering and leaving the specific _window_, not just the specific buffer.

This is now fixed by deleting the keymaps within the already existing `set_callback_for_reviewer_leave` function.

lua/gitlab/state.lua
```diff
@@ -232,6 +399,54 @@ M.merge_settings = function(args)
     )
   end
 
+  M._get_nested_field = function(table, field)
```
harrisoncramer, line 402: Let's pull this into the utils file, it could be generally useful!

lua/gitlab/state.lua
```diff
@@ -232,6 +399,54 @@ M.merge_settings = function(args)
...
+    "discussion_tree.toggle_unresolved_discussions",
+    "help",
+    "popup.keymaps.next_field",
+    "popup.keymaps.prev_field",
+    "popup.perform_action",
+    "popup.perform_linewise_action",
+  }
+  for _, field in ipairs(replaced) do
+    if M._get_nested_field(M.settings, field) ~= nil then
+      u.notify(
```
harrisoncramer, line 443: Rather than notifying them of every single field, let's aggregate these into a single message and just warn them once. You can also add the review pane check for Delta to this and make it slightly more generic, and perhaps name this "deprecated_settings_fields" or something like that to be used in the future.

lua/gitlab/state.lua
```diff
@@ -202,6 +240,135 @@ M.unlinked_discussion_tree = {
...
+    return
+  end
+
+  if keymaps.global.start_review then
+    vim.keymap.set("n", keymaps.global.start_review, function()
+      require("gitlab").review()
+    end, { desc = "Start Gitlab review", nowait = keymaps.global.start_review_nowait })
+  end
+
+  if keymaps.global.choose_merge_request then
```
harrisoncramer, line 257: Agreed that this can be dried up but no need to do that now, it's not really too broken, would rather have some repetition here than some fancy abstraction that makes it hard to change these in the future.

lua/gitlab/state.lua
```diff
@@ -60,14 +60,71 @@ M.settings = {
...
+      toggle_discussions = "gld",
+      toggle_draft_mode = "glD",
+      publish_all_drafts = "glP",
+    },
+    popup = {
+      disable_all = false,
       next_field = "<Tab>",
       prev_field = "<S-Tab>",
+      perform_action = "ZZ",
+      perform_linewise_action = "<leader>l",
```
harrisoncramer, line 94: If we are not using leaders universally this should match the other "action" keymaps, so probably ZL would work? I find it odd this is the only keymap prefixed with the leader key

doc/gitlab.nvim.txt
```diff
@@ -565,36 +607,94 @@ reviewer when checking out a new branch:
...
+keybinding: `cc` and `ss`. Alternatively, you can use any motion that moves on
+that line only, e.g., `cw`, `cl`, `c$`, etc.
+
+The plugin takes advantage of the fact that the `c` and `s` normal mode
+commands modify the text, but the Diffview buffers are set as |nomodifiable|,
+so the keys are practically free for remapping. If, however, you don't like
+the default `c` and `s` keybindings, you can remap these operators to a
+different key or keys, e.g. `keymaps.reviewer.create_comment = "<leader>c"`.
+This will still work for commenting on the same line by duplicating the
+keybinding, e.g., `<leader>c<leader>c`.
```
harrisoncramer, line 674: I'd delete this paragraph, this is explained in the previous paragraph. Or you could just add: "This same logic will apply if you remap this keybinding."

harrisoncramer (changes requested):
This looks great, thank you for working on it and sorry it took me so long to get to the review. I've left some suggestions but mostly this looks great to me! 🚀

jakubbortlik:
> This looks great, thank you for working on it and sorry it took me so long to get to the review. I've left some suggestions but mostly this looks great to me! 🚀

Hi. Thanks for the review. I'm on vacation and I'll get back to this next week.

lua/gitlab/state.lua
```diff
@@ -232,6 +399,54 @@ M.merge_settings = function(args)
...
+    "help",
+    "popup.keymaps.next_field",
+    "popup.keymaps.prev_field",
+    "popup.perform_action",
+    "popup.perform_linewise_action",
+  }
+  for _, field in ipairs(replaced) do
+    if M._get_nested_field(M.settings, field) ~= nil then
+      u.notify(
+        "The '" .. field .. "' setup option has been removed. Use the 'kyemaps' field instead.",
```
chardskarth, line 444:
```suggestion
        "The '" .. field .. "' setup option has been removed. Use the 'keymaps' field instead.",
```

doc/gitlab.nvim.txt
jakubbortlik: Agree, this is unnecessary detail. Removed.

lua/gitlab/state.lua
jakubbortlik: I've changed that to "ZA" because that's easier to type than "ZL" on qwerty or colemak (no change of hands). What do you think?

jakubbortlik: There is still `<leader>p` for printing a discussion node, but that's for debugging, so I believe it can stay that way.

lua/gitlab/state.lua
jakubbortlik: Good idea. I've changed it to just print once.

lua/gitlab/state.lua
jakubbortlik: Done.

lua/gitlab/state.lua
jakubbortlik: Maybe in the future it would be nice to add a proper health check and a proper config validation (something like suggested in the [PR for the plugin guideline](https://github.com/neovim/neovim/pull/29073)).

doc/gitlab.nvim.txt
```diff
@@ -565,36 +607,94 @@ reviewer when checking out a new branch:
...
+don't really care about your `co` mapping for this buffer, you can force the
+`gitlab.nvim` mapping to fire immediately by adding a special field into your
+`keymaps` table. E.g., if the name of the keybinding is
+`reviewer.create_comment`, then you can add the following:
+>lua
+    keymaps = { reviewer = { create_comment_nowait = true }}
+<
+This works by adding the |nowait| argument to the |vim.keymap.set| call.
+
+
```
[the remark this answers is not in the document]
jakubbortlik: I actually find the possibility to use `nowait` useful, so I'd be grateful if it could stay there. E.g., I use [debugprint](https://github.com/andrewferrier/debugprint.nvim/) and don't want to change it's default `g?` prefix for keybindings, but at the same time, I don't want to change `g?` for help in `gitlab.nvim` and I don't want to wait for it to apply.

Other plugins also make it possible to adjust `nowait` on their keymaps, e.g., [neo-tree](https://github.com/nvim-neo-tree/neo-tree.nvim), or [diffview](https://github.com/sindrets/diffview.nvim). So this I don't think adding this possibility to `gitlab.nvim` is somehow strange.

But I agree that the description was too much, so I've simplified it.

harrisoncramer (approved):

doc/gitlab.nvim.txt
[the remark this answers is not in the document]
harrisoncramer: Thank you, the simpler description in the help document makes me feel better about this change. I was not familiar with other plugins' use of this feature.

PR #337, opened 2024-07-30
chore(docs): add keybindings in README for which-key users

chardskarth:
I added these for better DX, because I don't think it's possible for devs to remember all these available feature. 
With `which-key`, the user should be able to see what each keybinding does.

---
This plugin is awesome, and very helpful.

I sure hope a lot of people will contribute to your project @harrisoncramer !

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,43 @@ For a list of all these settings please run `:h gitlab.nvim.configuring-the-plug
 
 The plugin sets up a number of useful keybindings in the special buffers it creates, and some global keybindings as well. Refer to the relevant section of the manual `:h gitlab.nvim.keybindings` for more details.
 
+### Or if you are using [which-key](https://github.com/folke/which-key.nvim), here's what you can use:
+
+```lua
+local wk = require("which-key")
+local gitlab = require("gitlab")
+
+wk.register({
+    ["lb"] = { gitlab.choose_merge_request, "choose_merge_request" },
+    ["lr"] = { gitlab.review, "review" },
+    ["ls"] = { gitlab.summary, "summary" },
+    ["lA"] = { gitlab.approve, "approve" },
+    ["lR"] = { gitlab.revoke, "revoke" },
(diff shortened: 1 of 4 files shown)


harrisoncramer:
Hey thank you very much for the MR! We are actually in the process of pulling these keybindings into the plugin itself, (see #331) so it probably makes sense for us to get that merged first, and then you can work from there? Some of the keybindings will change.

Thanks for the kind words!

chardskarth:
Marking this as draft until #331 gets done. Thanks!

harrisoncramer:
Hey @chardskarth I've merged in the keybindings changes into develop, if you want to give a quick look them now and adjust this, they should be ready. 

We can merge and ship this them both to main at the same time. Thanks!

README.md
```diff
@@ -154,6 +154,43 @@ vim.keymap.set("n", "glP", gitlab.publish_all_drafts)
 vim.keymap.set("n", "glD", gitlab.toggle_draft_mode)
 ```
 
+### Or if you are using [which-key](https://github.com/folke/which-key.nvim), here's what you can use:
```
harrisoncramer, line 157: This should go in `doc/gitlab.nvim.txt` probably at the bottom of the keybindings section.

README.md
```diff
@@ -154,6 +154,43 @@ vim.keymap.set("n", "glP", gitlab.publish_all_drafts)
...
+
+```lua
+local wk = require("which-key")
+local gitlab = require("gitlab")
+
+wk.register({
```
jakubbortlik, line 163: Hi, @chardskarth, please have a look at recent which-key's documentation. `wk.register` is deprecated ("Use `require("which-key").add()` instead.")

chardskarth:
Given some thought. I think the default keybinding works best. I think it's enough that the user can opt out from the default bindings. And if the user has that skill, they probably can spin up their own keybinding anyway.

Thanks for taking time on this. Closing now.

PR #339, opened 2024-08-07
Added posibility to change discussion-tree chevrons

c0r73x:

PR #340, opened 2024-08-08
Release

harrisoncramer:
feat!: Implements default keybindings for all actions. This is a breaking change to the way the plugin is configured. If users are using the old configuration the plugin will warn them which fields have been removed in their configuration. Old keybindings can be found here: https://github.com/harrisoncramer/gitlab.nvim/pull/340#issuecomment-2282756924 (#331)
feat: Customize discussion tree chevrons (#339)

This is a #MAJOR release.


jakubbortlik:
> feat!: Implements default keybindings for all actions. This is a breaking change to the way the plugin is configured. If users are using the old configuration the plugin will warn them which fields are deprecated in their configuration (#331) feat: Customize discussion tree chevrons (#339)
> 
> This is a #MAJOR release.

Hi, I would suggest changing "deprecated" to "removed", otherwise it sounds like the fields can still be used, and only will be removed in the future.

harrisoncramer:
Hey @jakubbortlik I've been using the develop branch locally and have noticed a lot of my muscle memory has been changed by the key-maps we updated. 

To help impacted users, can we please post a comment here with an update of what the previous keybindings were, and what they've changed to, so that users like myself can easily add this to their settings? For instance, finish this config, please!

```lua
gitlab.setup({
      keymaps = {
        global = {
          start_review = "glr", -- Old keybinding
          start_review_nowait = true,
        },
        discussion_tree = {
          switch_view = "S",
          switch_view_nowait = true,
        },
      },
    })
```

I can link it in the commit message itself so folks will see.

jakubbortlik:
> Hey @jakubbortlik I've been using the develop branch locally and have noticed a lot of my muscle memory has been changed by the key-maps we updated.
> 
> To help impacted users, can we please post a comment here with an update of what the previous default keybindings were, so that users like myself can easily add this to their settings? For instance, finish this config, please!
> 
> ```lua
> gitlab.setup({
>       keymaps = {
>         global = {
>           start_review = "glr", -- Old keybinding
>           start_review_nowait = true,
>         },
>         discussion_tree = {
>           switch_view = "S",
>           switch_view_nowait = true,
>         },
>       },
>     })
> ```
> 
> I can link it in the commit message itself so folks will see -- just trying to be as careful as possible to reduce impact on our users.

Hi, that's a good a idea. I'm away from my computer now, can do it on Monday.

jakubbortlik:
Hi @harrisoncramer, here is the configuration that will more or less restore the original keybindings. The exception is that `glc` does not create a single line comment in normal mode as it used to, because all comment creation is now realized with just `create_multiline_comment`, so the user either has to create a visual selection first or follow `glc` with a motion command.
I've also added some notes, mostly on why I think the new keybindings are superior :)

I hope this is what you were asking for:

```lua
gitlab.setup({
  keymaps = {
    global = {
      create_mr = "glO",
      choose_merge_request = "glb",
      start_review = "glr", -- Don't use `start_review_nowait = true` with this setting, otherwise `glra` and `glrd` for managing reviewers will not be available.
    },
    popup = {
      perform_action = "<leader>s",
      perform_linewise_action = "<leader>l",
    },
    discussion_tree = {
      toggle_resolved = "p",
      jump_to_reviewer = "m", -- Since gitlab.nvim's discussion tree buffer is non-modifiable, you can use a modifying key instead ("a", "x", etc.) and not lose the "m" functionality for mark setting.
      switch_view = "S",
      refresh_data = "a",
    },
    reviewer = { -- Note that these are not global but buffer-local keybindings, so the `gl` prefix is rather redundant.
      create_comment = "glc", -- Runs `gitlab.create_multiline_comment` in visual & normal mode. In normal mode must be followed by a motion, e.g., "$" to work on a single line.
      create_suggestion = "glC", -- Runs `gitlab.create_comment_suggestion` in visual & normal mode. In normal mode must be followed by a motion.
      move_to_discussion_tree = "glm", -- You'll probably find it more practical to use the same keybinding as for `discussion_tree.jump_to_reviewer`.
    }
  }
})
```

PR #342, opened 2024-08-09
Fix: Discussion tree highlighting

jakubbortlik:
#339 introduced some regressions in discussion tree highlighting. This PR fixes:
1. expander highlighting in sub-nodes
2. highlighting of usernames with dots (e.g., `@name.surname`)
3. separate highlighting for mentions (different from `GitlabUsername`)


harrisoncramer:
🚀 🚀 🚀

PR #344, opened 2024-08-12
Update README.md

briandipalma:
Also `review` doesn't have a keybinding. For my workflow I never need to choose a PR to review as I may be reviewing 2-5 PRs at the same time so they are all open in their own terminal tab/worktree directory so I just want to open what I'm in right now and go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ To view these help docs and to get more detailed help information, please run `:
 
 1. Install Go
 2. Add configuration (see Installation section)
-5. Run `:lua require("gitlab").choose_merge_request()`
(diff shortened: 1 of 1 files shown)


jakubbortlik:
Hi Brian. The keybinding for `review` is `global.start_review`, which is bound to `glS` by default. I've made the keybinding field a little more explicit than just `review`, at the expense of inconsistency with the function name (maybe I should have renamed the function as well, with a deprecated alias). I've now updated the [note](https://github.com/harrisoncramer/gitlab.nvim/pull/340#issuecomment-2279646521) about how to recreate the original keybindings, so that it explicitly mentions `gitlab.review()`.

PR #345, opened 2024-08-14
feat: Add healthcheck

harrisoncramer:
Adds health check to plugin, and adds types for all settings.


harrisoncramer:
@jakubbortlik Take a look at this when you have a moment. It sets up types for the plugin so folks can get a nice autocomplete and warnings when configuring it, and adds a healthcheck as well!

PR #351, opened 2024-08-19
Fix: Error with removed config fields

jakubbortlik:
Hi @harrisoncramer, I didn't have the time to react to your [comment](https://github.com/harrisoncramer/gitlab.nvim/pull/345#issuecomment-2291539147) on your initial MR for the health check. Now that I've given it a try, I've found a bug in the check for removed config fields, so here's a fix.


harrisoncramer:
Ah, thanks good catch. Sorry about that!

PR #352, opened 2024-08-23
Feat: Enable always jumping to discussion tree

### LINKED ISSUE #309: Fallback for gitlab.move_to_discussion_tree_from_diagnostic
## Feature Description

In my `gitlab.nvim` config I've mapped both `gitlab.move_to_discussion_tree_from_diagnostic` and `jump_to_reviewer` to the same key (`a`) and I use it to quickly jump between the discussion tree and the reviewer. This workflow is interrupted when jumping to the reviewer for a resolved thread with `settings.discussion_signs.skip_resolved_discussion = true`. Then I just get the warning "No diagnostics for this line".

I'd like the `gitlab.move_to_discussion_tree_from_diagnostic` function to allow a fallback for such cases. Instead of just showing a warning, I'd like focus to be brought to the last position in the discussion tree.

I imagine it should be fairly easy to implement as an opt-in feature, so the default behaviour would not change. I can open a PR after the other pending PR are sorted out.

jakubbortlik:
This PR makes it possible to customize the behavior when pressing the `move_to_discussion_tree` keybinding in the reviewer and there is no diagnostic under the cursor. Users can now set up any combination of:
- show warning
- jump to last position in discussion tree

Closes #309.

### DIFF
--- a/lua/gitlab/actions/discussions/init.lua
+++ b/lua/gitlab/actions/discussions/init.lua
@@ -202,14 +202,11 @@ M.move_to_discussion_tree = function()
   end
 
   if #d == 0 then
-    local warning = "No diagnostics for this line."
-    if state.settings.reviewer_settings.no_diagnostic_actions.jump then
-      warning = warning .. " Jumping to last position in discussion tree."
+    if state.settings.reviewer_settings.jump_with_no_diagnostics then
       vim.api.nvim_win_set_cursor(M.split.winid, { M.last_row, M.last_column })
       vim.api.nvim_set_current_win(M.split.winid)
-    end
-    if state.settings.reviewer_settings.no_diagnostic_actions.warn then
-      u.notify(warning, vim.log.levels.WARN)
+    else
+      u.notify("No diagnostics for this line.", vim.log.levels.WARN)
     end
     return
   elseif #d > 1 then
--- a/lua/gitlab/actions/discussions/init.lua
+++ b/lua/gitlab/actions/discussions/init.lua
@@ -202,7 +202,15 @@ M.move_to_discussion_tree = function()
   end
 
   if #d == 0 then
-    u.notify("No diagnostics for this line", vim.log.levels.WARN)
+    local warning = "No diagnostics for this line."
+    if state.settings.reviewer_settings.no_diagnostic_actions.jump then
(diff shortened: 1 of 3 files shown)


jakubbortlik:
Hi @harrisoncramer, this PR changes the current behaviour - the new default is to always jump to the discussion tree even if there is no diagnostic under the cursor, as I believe this behaviour will be preferred by the majority of users. If, instead, you'd like the default to be that only a warning is shown (same as the old behaviour), I'll make the necessary changes.

lua/gitlab/actions/discussions/init.lua
```diff
@@ -202,7 +202,15 @@ M.move_to_discussion_tree = function()
...
   if #d == 0 then
-    u.notify("No diagnostics for this line", vim.log.levels.WARN)
+    local warning = "No diagnostics for this line."
+    if state.settings.reviewer_settings.no_diagnostic_actions.jump then
+      warning = warning .. " Jumping to last position in discussion tree."
+      vim.api.nvim_win_set_cursor(M.split.winid, { M.last_row, M.last_column })
+      vim.api.nvim_set_current_win(M.split.winid)
+    end
+    if state.settings.reviewer_settings.no_diagnostic_actions.warn then
+      u.notify(warning, vim.log.levels.WARN)
```
harrisoncramer, line 212: This is a pretty simple feature, and I don't think it merits two new fields in the configuration table. I'd like to keep the API surface as small as possible for users.

Let's have it be a single option called `jump_with_no_diagnostics` and if it's enabled, do not show any warning and do the jump. Otherwise (it should be the. default case) do not jump and show the warning. 

```lua
    if state.settings.reviewer_settings.jump_with_no_diagnostics then
      vim.api.nvim_win_set_cursor(M.split.winid, { M.last_row, M.last_column })
      vim.api.nvim_set_current_win(M.split.winid)
    else
      local warning = "No diagnostics for this line."
      u.notify(warning, vim.log.levels.WARN)
    end
```

harrisoncramer:
Small change, but once you make that I'll merge this in. Sorry this took me so long to get to.

lua/gitlab/actions/discussions/init.lua
jakubbortlik: OK, fixed. Thanks for the review.

PR #355, opened 2024-09-05
Make help popup not editable and close it on BufLeave

EpiCanard:
Hello,

This PR make the help popup not editable by calling switch_can_edit_buf with false.
It also add a popup listener to automatically close the help popup when the buffer is leaved.


harrisoncramer:
Thank you for the contribution! Looks good!

Issue #319, opened 2024-06-20
choose_merge_request does not select correct MR if there are several MRs into different target branches


jakubbortlik:
## Prerequisites

- [x] I'm on the latest version of the plugin
- [x] I've installed the required dependencies
- [x] I've run `:h gitlab.nvim.troubleshooting` and followed the steps there

## Setup Configuration and Environment

n/a

## Bug Description

The `choose_merge_request` action only chooses the source branch, but it always selects the same MR, even if I select different MRs that have different target branches.

I admit that this is a somewhat artificial use case -- I've never really seen MRs for the same feature branch but into different target branches, but I use this scenario in testing `gitlab.nvim` and some users might actually encounter this in real work.

## Reproduction Steps

1. Create a feature branch
2. Create multiple MRs into different target branches
3. Run `choose_merge_request` for all the MRs

## Expected behaviour

When I choose a MR, the correct feature as well as target branch gets selected.

PR #357, opened 2024-09-06
fix: Fixes MR selection if there are multiple targets

harrisoncramer:
Addresses #319

### DIFF
--- a/cmd/app/config.go
+++ b/cmd/app/config.go
@@ -9,6 +9,7 @@ type PluginOptions struct {
 		Request  bool `json:"go_request"`
 		Response bool `json:"go_response"`
 	} `json:"debug"`
+	ChosenTargetBranch *string `json:"chosen_target_branch,omitempty"`
 	ConnectionSettings struct {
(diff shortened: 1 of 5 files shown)


harrisoncramer:
Hey @jakubbortlik I've fixed that issue.

If you are just opening the reviewer in a merge request and there are multiple options, I'll throw an error in the Go server and tell you to choose one via the `choose_merge_request` option. The `choose_merge_request` function now sets a value in the shared state that specifies the target for the MR selector, which is passed along to the server.

This should resolve your issue, can you please give it a try?

jakubbortlik:
Hi Harrison, I confirm that in this branch the issue is solved. Thanks!

jakubbortlik:
Maybe you could show the error and run `choose_merge_request` right away. I believe that's what most users will want to do anyway.

harrisoncramer:
Yeah, I thought about having that, but it would add some additional complexity to the codebase for an edge case that is very rarely going to occur, so I don't think it's worth it 👍

PR #369, opened 2024-09-13
fix: Add better checks for leaving comments

harrisoncramer:
Stricter checks for when someone attempts to leave a comment in the wrong place. We are now checking for attempted comments in the file tree (wrong) and the discussion tree (also wrong). We are now not opening the layout entirely, instead of erroring on the actual attempt to create a comment.

You can replicate this error if you call `require("gitlab").create_comment()` while hovering over either of those locations.

### DIFF
--- a/lua/gitlab/reviewer/init.lua
+++ b/lua/gitlab/reviewer/init.lua
@@ -151,25 +151,7 @@ end
 ---other modules such as the comment module to create line codes or set diagnostics
 ---@return DiffviewInfo | nil
 M.get_reviewer_data = function()
-  if M.tabnr == nil then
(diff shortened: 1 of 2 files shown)


lua/gitlab/reviewer/init.lua
```diff
@@ -151,25 +151,7 @@ end
...
-    return
-  end
-
-  -- Check if we are in the diffview buffer
   local view = diffview_lib.get_current_view()
-  if view == nil then
-    u.notify("Could not find Diffview view", vim.log.levels.ERROR)
-    return
-  end
-
```
harrisoncramer, line 172: These were moved up so that we error earlier, so that folks are blocked from even opening the popup in the first place.

PR #380, opened 2024-09-19
Feat: Automatically open fold under cursor

jakubbortlik:
When jumping to the reviewer for a comment on an unchanged line, that line may be in a fold which DiffView closes every time. After jumping to that note a couple of times, this is a little annoying, so in this MR, I have the fold opened automatically.

### DIFF
--- a/lua/gitlab/reviewer/init.lua
+++ b/lua/gitlab/reviewer/init.lua
@@ -145,6 +145,7 @@ M.jump = function(file_name, line_number, new_buffer)
     line_number = number_of_lines
   end
   vim.api.nvim_win_set_cursor(0, { line_number, 0 })
+  u.open_fold_under_cursor()
 end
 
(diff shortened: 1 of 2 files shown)


lua/gitlab/reviewer/init.lua
```diff
@@ -145,6 +145,7 @@ M.jump = function(file_name, line_number, new_buffer)
     line_number = number_of_lines
   end
   vim.api.nvim_win_set_cursor(0, { line_number, 0 })
+  u.open_fold_under_cursor()
```
jakubbortlik, line 148: I'm also thinking about adding the following line, to redraw with the current line in the center of the window. It can make more context visible:
```lua
  vim.cmd("normal! zz")
```

jakubbortlik: OK, I've just added that too, because I find myself always centering the window manually and not enjoying it at all :)

PR #395, opened 2024-10-11
fix: Incorrect warning about commits

harrisoncramer:
This MR updates the function that checks the stability of the branch and MR.

There is now one function responsible for checking whether the current branch is ahead/behind it's tracking branch. That function will warn the user if that's the case for all three possibilites (ahead, behind, or ahead and behind).

The second function is responsible for alerting the user if the status of the current MR is bad: merged, closed, or with conflicts.

These should address the points outlined in the issue outlined by Jakub. For instance, in the complex case where the user is setup to track a remote and they are behind the remote, and the MR associated with that remote has already been closed, they will get both warnings, which is accurate.

Plesae @jakubbortlik give this a look when you have time. Thanks.

### DIFF
--- a/lua/gitlab/actions/summary.lua
+++ b/lua/gitlab/actions/summary.lua
@@ -82,7 +82,7 @@ M.summary = function()
     vim.api.nvim_set_current_buf(description_popup.bufnr)
   end)
 
-  git.current_branch_up_to_date_on_remote(vim.log.levels.WARN)
+  git.check_current_branch_up_to_date_on_remote(vim.log.levels.WARN)
   git.check_mr_in_good_condition()
 end
 
--- a/lua/gitlab/actions/summary.lua
+++ b/lua/gitlab/actions/summary.lua
@@ -83,6 +83,7 @@ M.summary = function()
   end)
 
   git.current_branch_up_to_date_on_remote(vim.log.levels.WARN)
+  git.check_mr_in_good_condition()
 end
 
 -- Builds a lua list of strings that contain metadata about the current MR. Only builds the
--- a/lua/gitlab/git.lua
+++ b/lua/gitlab/git.lua
@@ -150,7 +150,7 @@ end
 ---Returns true if `branch` is up-to-date on remote, otherwise false and warns user
 ---@param log_level integer
 ---@return boolean
-M.current_branch_up_to_date_on_remote = function(log_level)
+M.check_current_branch_up_to_date_on_remote = function(log_level)
   local u = require("gitlab.utils")
 
   -- Get current branch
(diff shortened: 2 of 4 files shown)


lua/gitlab/actions/summary.lua
```diff
@@ -83,6 +83,7 @@ M.summary = function()
   end)
 
   git.current_branch_up_to_date_on_remote(vim.log.levels.WARN)
+  git.check_mr_in_good_condition()
```
harrisoncramer, line 86: It's nice to have this warning here as well, in case the user is just looking at the summary outside of the reviewer pane.

lua/gitlab/git.lua
```diff
@@ -43,6 +43,60 @@ M.switch_branch = function(branch)
...
 
+---Fetches the name of the remote tracking branch for the current branch
+---@return string|nil, string|nil
+M.get_remote_branch = function()
+  return run_system({ "git", "rev-parse", "--abbrev-ref", "--symbolic-full-name", "@{u}" })
+end
+
+---Determines whether the tracking branch is ahead of or behind the current branch, and warns the user if so
+---@param current_branch string
+---@param remote_branch string
```
jakubbortlik, line 54:
```suggestion
---@param remote_branch string
---@param log_level integer
```

lua/gitlab/git.lua
```diff
@@ -93,39 +147,45 @@ M.contains_branch = function(current_branch)
   return run_system({ "git", "branch", "-r", "--contains", current_branch })
 end
 
----Returns true if `branch` is up-to-date on remote, false otherwise.
+---Returns true if `branch` is up-to-date on remote, otherwise false and warns user
 ---@param log_level integer
----@return boolean|nil
+---@return boolean
 M.current_branch_up_to_date_on_remote = function(log_level)
```
jakubbortlik, line 153: To make it consistent with `check_mr_in_good_condition`:
```suggestion
M.check_current_branch_up_to_date_on_remote = function(log_level)
```

lua/gitlab/git.lua
```diff
@@ -43,6 +43,60 @@ M.switch_branch = function(branch)
...
+    u.notify("Error parsing ahead/behind information.", vim.log.levels.ERROR)
+    return false
+  end
+
+  ahead = tonumber(ahead)
+  behind = tonumber(behind)
+
+  if ahead > 0 and behind == 0 then
+    u.notify(
+      string.format("You have local commits that are not on %s. Have you forgotten to push?", remote_branch),
```
jakubbortlik, line 76: Maybe you could make this consistent with the other warning:
```suggestion
      string.format("There are local changes that haven't been pushed to %s yet", remote_branch),
```

jakubbortlik (commented):
Hi Harrison, this looks much more robust than the original way of checking for branch state. I have three small suggestions that you may consider. I've tested the basic cases and all worked as expected. Thanks!

PR #397, opened 2024-10-12
fix: choose correct MR if multiple are from same source branch

harrisoncramer:
We'll use the internal ID of the Merge Request rather than the source/target branch when a specific MR is chosen, which is guaranteed to be unique and will avoid situations where we pass along a source/target relationship that is not unique. This is unusual but technically still possible if someone is repeatedly opening MRs against a branch (say main) from another branch, merging them and deleting the source branch, then re-creating a new MR with the same branch.


harrisoncramer:
Moved to https://github.com/harrisoncramer/gitlab.nvim/pull/398

Issue #385, opened 2024-09-30
Use conceal functionality to show emojis (:thumbsup: -> 👍)


jakubbortlik:
## Feature Description
I think it would be quite nice to use the conceal functionality  in markdown buffers like the discussion tree and the comment composition buffers to _visually_ replace the text like `:thumbsup:` with the actual Unicode symbols like :+1:.

@harrisoncramer, do you think this is something you'd be willing to include in the plugin? I haven't checked whether there is another plugin that does this.

I'm only afraid that adding 570 or more syntax rules for each emoji such as `syntax match emoji_tada ":tada:" conceal cchar=🎉` could negatively affect responsiveness of the UI.

harrisoncramer:
If we wanted to add something like this it'd probably be worth making opt-in. I'm not sure what the performance implications would be, you're welcome to give it a try and see. Other Gitlab integrations (like IntelliJ has one) don't do this

jakubbortlik:
From the user perspective, this is basically just a cosmetic thing, so if you don't mind keeping the issue open foe some time, I'll get back to it after some more functional issues are sorted out. I've got a couple ideas about emojis, so I'll just list them here, but I can open separate issues, if you prefer.

1. Where does the list of emojis in `config/emojis.json` come from? Gitlab online seems to support quite a few more than `gitlab.nvim` does. I'll be adding those that come up in our team's discussions, but maybe I could try to find some more exhaustive list from Gitlab right away.
2. The emojis in `config/emojis.json` also include a `short_name` (the string that one would normally have to type literally in markdown for the emoji to show up on Gitlab) and an array of `keywords`. It would be great, if these were also used for fuzzy searching for an emoji.
3. It would be nice if the emojis could be autocompleted while working on markdown buffers, but I guess there are existing plugins for this, so it's probably superfluous for `gitlab.nvim` to implement this functionality.

jakubbortlik:
> From the user perspective, this is basically just a cosmetic thing, so if you don't mind keeping the issue open foe some time, I'll get back to it after some more functional issues are sorted out. I've got a couple ideas about emojis, so I'll just list them here, but I can open separate issues, if you prefer.
> 
>     1. Where does the list of emojis in `config/emojis.json` come from? Gitlab online seems to support quite a few more than `gitlab.nvim` does. I'll be adding those that come up in our team's discussions, but maybe I could try to find some more exhaustive list from Gitlab right away.
> 
[...]

PR #405, opened 2024-10-15
Fix: Use correct name for emoji

jakubbortlik:
When updating the `emojis.json` I made a mistake for the :tada: emoji. The emoji could be selected and awarded, but it was not shown in the `gitlab.nvim` UI.

This PR fixes it.


jakubbortlik:
I'm merging this in order to be able to work on #385 without introducing conflicts.

PR #406, opened 2024-10-15
Fix: Restore cursor when updating from outside of tree

jakubbortlik:
This PR improves the user experience in the discussion tree by restoring cursor position after rebuilding the tree. This makes navigation in the discussion tree more fluent without the cursor jumping to the first line of the tree after actions like resolving discussions, adding new comments, publishing drafts etc.

### DIFF
--- a/lua/gitlab/actions/comment.lua
+++ b/lua/gitlab/actions/comment.lua
@@ -63,8 +63,6 @@ local confirm_create_comment = function(text, visual_range, unlinked, discussion
     return
   end
 
-  vim.print("Here: ", unlinked, discussion_id)
-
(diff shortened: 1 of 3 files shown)


jakubbortlik:
I've marked this a draft because I've found a bug that this PR introduces. When data are refreshed while the discussion tree is closed, there is an error when trying to reopen the discussion tree again. I'm working on a fix.

jakubbortlik:
I've now fixed the code so that it works correctly for both discussion trees even if the cursor is outside of the discussion tree.

There are some problems with test (`Failed to install LuaRocks`) that I don't think I can do anything about and would appreciate if you @harrisoncramer could fix if you find this PR useful.

PR #416, opened 2024-11-05
fix: comment creation should not be possible for renamed and moved files

harrisoncramer:
Makes it not possible to leave comments on renamed files

### DIFF
--- a/lua/gitlab/hunks.lua
+++ b/lua/gitlab/hunks.lua
@@ -90,18 +90,28 @@ local line_was_added = function(linnr, hunk, all_diff_output)
 end
 
 ---Parse git diff hunks.
----@param file_path string Path to file.
 ---@param base_sha string Git base SHA of merge request.
 ---@return HunksAndDiff
-local parse_hunks_and_diff = function(file_path, base_sha)
+local parse_hunks_and_diff = function(base_sha)
   local hunks = {}
   local all_diff_output = {}
 
-  local Job = require("plenary.job")
+  local reviewer = require("gitlab.reviewer")
+  local cmd = {
+    "diff",
+    "--minimal",
+    "--unified=0",
+    "--no-color",
+    base_sha,
+    "--",
+    reviewer.get_current_file_oldpath(),
+    reviewer.get_current_file_path(),
+  }
 
+  local Job = require("plenary.job")
   local diff_job = Job:new({
     command = "git",
-    args = { "diff", "--minimal", "--unified=0", "--no-color", base_sha, "--", file_path },
+    args = cmd,
     on_exit = function(j, return_code)
       if return_code == 0 then
         all_diff_output = j:result()
@@ -139,8 +149,7 @@ end
 
 --- Processes the number of changes until the target is reached. This returns
 --- a negative or positive number indicating the number of lines in the hunk
---that have been added or removed prior to the target line
----comment
+--- that have been added or removed prior to the target line
 ---@param line_number number
 ---@param hunk Hunk
 ---@param lines table
@@ -221,11 +230,10 @@ end
 ---This is in order to build the payload for Gitlab correctly by setting the old line and new line.
 ---@param old_line number|nil
 ---@param new_line number|nil
----@param current_file string
 ---@param is_current_sha_focused boolean
 ---@return string|nil
-function M.get_modification_type(old_line, new_line, current_file, is_current_sha_focused)
-  local hunk_and_diff_data = parse_hunks_and_diff(current_file, state.INFO.diff_refs.base_sha)
+function M.get_modification_type(old_line, new_line, is_current_sha_focused)
+  local hunk_and_diff_data = parse_hunks_and_diff(state.INFO.diff_refs.base_sha)
   if hunk_and_diff_data.hunks == nil then
     return
   end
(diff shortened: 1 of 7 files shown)


lua/gitlab/hunks.lua
```diff
@@ -90,18 +90,28 @@ local line_was_added = function(linnr, hunk, all_diff_output)
...
-  local Job = require("plenary.job")
+  local reviewer = require("gitlab.reviewer")
+  local cmd = {
+    "diff",
+    "--minimal",
+    "--unified=0",
+    "--no-color",
+    base_sha,
+    "--",
+    reviewer.get_current_file_oldpath(),
```
harrisoncramer, line 107: Include the old path when calculating the diff, since the new file may have a different path

jakubbortlik:
Hi Harrison, thanks for doing this! It works in that commenting on renamed files is disabled, but for some default commenting keybindings, more errors are shown than intended:


The normal mode keybinding `ss` and the visual mode keybinding `s` show this error:
```
Error executing vim.schedule lua callback: ....../gitlab.nvim/lua/gitlab/actions/comment.lua:365: attempt to index field 'comment_popup' (a nil value)
stack traceback:                                                                                                                                                                                                                                                                                                                
        ....../gitlab.nvim/lua/gitlab/actions/comment.lua:365: in function <....../gitlab.nvim/lua/gitlab/actions/comment.lua:363>                                                                                                                                                                          
Press ENTER or type command to continue
```
This can be fixed by returning early from `create_comment_suggestion`:
```diff
M lua/gitlab/actions/comment.lua
@@ -359,6 +359,8 @@ M.create_comment_suggestion = function()
   local layout = M.create_comment_layout({ ranged = range_length > 0, unlinked = false })
   if layout ~= nil then
     layout:mount()
+  else
+    return
   end
   vim.schedule(function()
     if suggestion_lines then
```

The keybindings `cc`, `ss` for commenting/making a suggestion on the current line and combinations `c<motion>` and `s<motion` also show this message:
```
E5108: Error executing lua ....../gitlab.nvim/lua/gitlab/reviewer/init.lua:292: Vim:gitlab.nvim: Commenting on (unchanged) renamed or moved files is not supported                                                                                                                                                  
stack traceback:                                                                                                                                                                                                                                                                                                                
        [C]: in function 'nvim_cmd'                                                                                                                                                                                                                                                                                             
        ....../gitlab.nvim/lua/gitlab/reviewer/init.lua:292: in function <....../gitlab.nvim/lua/gitlab/reviewer/init.lua:290>                                                                                                                                                                          
Press ENTER or type command to continue
```
This will require some fixes of how the callback is executed for the operator mappings. I haven't found a trivial fix when briefly investigating this.

Also, interestingly when the `cc` mapping fails, nvim stays in visual mode, while for `ss` it stays in gets back to normal mode but cursor moves to the end of the intermediate visual selection. This should be fixed so that ideally the visual mode is never even entered (`lua/gitlab/reviewer/init.lua` line  291: `vim.api.nvim_cmd({ cmd = "normal", bang = true, args = { "'[V']" } }, {})`) with these mappings.

harrisoncramer:
I've fixed the first issue you pointed out, using your provided solution.

I'm not seeing the second issue at all, ranged comments or suggestions are just giving me the single error message.

jakubbortlik:
> I'm not seeing the second issue at all, ranged comments or suggestions are just giving me the single error message.

Interesting. I'll find out what the problem is and let you know.

jakubbortlik:
Hi @harrisoncramer, you should be able to reproduce the issue with this config:
```lua
-- This is a minimal init.lua that is used for debugging
local root = vim.fn.stdpath("run") .. "/nvim/mininit"
vim.fn.mkdir(root, "p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  {
    "harrisoncramer/gitlab.nvim",
    dir = "~/projects/gitlab.nvim/",
    dependencies = {
      "MunifTanjim/nui.nvim",
      "nvim-lua/plenary.nvim",
      "sindrets/diffview.nvim",
      "stevearc/dressing.nvim",   -- Recommended but not required. Better UI for pickers.
      "nvim-tree/nvim-web-devicons" -- Recommended but not required. Icons in discussion tree.
    },
    enabled = true,
    build = function() require("gitlab.server").build(true) end, -- Builds the Go binary
    config = function()
      require("gitlab").setup({
        colors = {
          discussion_tree = {
            mention = "Keyword",  -- Overrides to get rid of errors due to missing colors...
            directory = "Keyword",
            resolved = "Keyword",
            unresolved = "Keyword",
            draft = "Keyword"
          },
        },
      })
    end,
  },
}

require("lazy").setup(plugins, {
  root = root .. "/plugins",
})
```

The way I understand this is that the problem is in lines 292-295 in `lua/gitlab/reviewer/init.lua`:
```lua
    vim.api.nvim_cmd(
      { cmd = "lua", args = { ("require'gitlab'.%s()"):format(callback) }, mods = { lockmarks = true } },
      {}
    )
```
The docs on `vim.api.nvim_cmd` say: `On execution error: fails with Vimscript error, updates v:errmsg.` I think line 185 in `lua/gitlab/actions/comment.lua`: `u.notify("Commenting on (unchanged) renamed or moved files is not supported", vim.log.levels.ERROR)` causes this execution error and hence the verbose error.

I've looked at your nvim config and you seem to be using `rcarriga/nvim-notify` which works around this problem.

jakubbortlik:
I still haven't figured out why for `cc` nvim stays in visual mode while for `ss` it turns back to normal mode.

PR #422, opened 2024-11-12
Feat: Enable sorting discussions by original comment

### LINKED ISSUE #408: Make sorting discussions customizable
## Feature Description
I would appreciate very much if I could configure the plugin to sort the discussions in the discussion tree by the date of the original comment (root node), rather than the date of the last reply in the thread. When discussions in the tree change position due to new replies, it makes it more difficult for me to orient myself in the tree.

This would require changes in the go server that I may not be able to do myself, even though I can give it a try.

jakubbortlik:
Resolves #408.

This PR modifies the Go server in order to add a different sorting order for discussions. Users can now choose between the original sorting order by `latest_reply`, in which the the recently active threads are on top of the tree and by `original_comment`, in which the threads are sorted chronologically from oldest on to to newest (just like on Gitlab online).

There're theoretically two more sorting orders, in which the order is reversed for the two existing options, by in my view they are not really practical so I haven't included them, since that would require a more complicated solution (the config `settings.discussion_tree.sort_by` would have to be a table, rather than a simple string. Of course, if you @harrisoncramer think that you'd actually want it to be more versatile and future-proof, I can change the PR so that the `sort_by` config option *is* a table.


cmd/app/list_discussions.go
```diff
@@ -30,20 +39,30 @@ type DiscussionsResponse struct {
...
+func (d SortableDiscussions) Less(i, j int) bool {
+	var iTime, jTime *time.Time
+	if d.SortBy == SortByOriginalComment {
+		iTime = d.Discussions[i].Notes[0].CreatedAt
+		jTime = d.Discussions[j].Notes[0].CreatedAt
+		return iTime.Before(*jTime)
+	} else { // SortByLatestReply
+		iTime = d.Discussions[i].Notes[len(d.Discussions[i].Notes)-1].CreatedAt
+		jTime = d.Discussions[j].Notes[len(d.Discussions[j].Notes)-1].CreatedAt
+		return iTime.After(*jTime)
```
harrisoncramer, line 60: Small nitpick, you don't need multiple returns here, you can just return outside of this if/else

harrisoncramer:
This sorting seems backward to me, when I sort I'd expect the newest comments to appear on top, here they appear on the bottom

lua/gitlab/actions/discussions/init.lua
```diff
@@ -795,6 +805,17 @@ M.toggle_draft_mode = function()
...
 end
 
+---Toggle between draft mode (comments posted as drafts) and live mode (comments are posted immediately)
+M.toggle_sort_by = function()
+  if state.settings.discussion_tree.sort_by == "original_comment" then
+    state.settings.discussion_tree.sort_by = "latest_reply"
+  else
+    state.settings.discussion_tree.sort_by = "original_comment"
+  end
+  u.notify("Sort discussion tree by '" .. state.settings.discussion_tree.sort_by .. "'", vim.log.levels.INFO)
```
harrisoncramer, line 815: Rather than having this be a notification, I think this information ought to live in the winbar!

harrisoncramer: I'm thinking just the arrow would work. I'd like to show this state to the user since it's persisted and affects the view of the discussion tree.

It's very possible we'll want to add filtering/searching by text to the winbar, so keeping the sort direction as short as possible is probably best to make room for that.

jakubbortlik: Actually, I'm addressing the winbar width in #435. Maybe have a look at that PR and let me know if you in general agree with the idea. Then we can find a "shrinkable" way for showing the sort order.

jakubbortlik: What about these variants of the description that would be selected based on available space:
1. `↑ by reply` -> `↑ reply` -> `↑`
2.  `↓ by thread` -> `↓ thread` -> `↓`
I think Unicode characters `U+2191` and `U+2193` should not be a problem for any user.

harrisoncramer: "↑ reply" and "↑ thread" look good to me! Clear and short.

jakubbortlik: I've removed the notification and made a couple of small improvements.

PR #426, opened 2024-11-19
Feat: Improve popup UX

jakubbortlik:
This PR attempts to improve user experience with popups in `gitlab.nvim` in several ways:

1. It makes it possible again to configure popup size (for comment popups this was disable by adding the `draft` field)
2. It limits the popups whose height is calculated from the content (help and pipeline) to the height of the window
3. It makes popups automatically resizable when the editor is resized (terminal zoomed in, tmux splits, moving to a smaller monitor)
4. It adds the autocommand for re-focusing last active window to all popups (it was only part of comment popups)
5. It improves visibility of the help popup
6. It makes the position of the popups customizable, which enables e.g. a setup where the comment popup doesn't obscure the reviewer

image


jakubbortlik:
Hi @harrisoncramer, I realize that there're many changes in this PR so I will really appreciate if you find the time to review and merge it. I probably should have split it into separate PRs, and I can still do this if you'd prefer. But at the same time, I've really taken care to make individual commits meaningful which should simplify review.

lua/gitlab/actions/summary.lua
```diff
@@ -166,15 +167,16 @@ M.edit_summary = function()
 end
 
 M.create_layout = function(info_lines)
-  local title_popup = Popup(u.create_box_popup_state(nil, false))
+  local settings = u.merge(state.settings.popup, state.settings.popup.summary or {})
+  local title_popup = Popup(u.create_box_popup_state(nil, false, settings))
   M.title_bufnr = title_popup.bufnr
-  local description_popup = Popup(u.create_box_popup_state("Description", true))
```
harrisoncramer, line 171: If we are splitting apart the popup functionality into it's own module that module should also contain `create_box_popup_state` and `create_popup_state`

lua/gitlab/state.lua
```diff
@@ -133,15 +133,16 @@ M.settings = {
...
     height = "60%",
+    position = "50%",
     border = "rounded",
     opacity = 1.0,
-    edit = nil,
     comment = nil,
+    edit = nil,
     note = nil,
-    help = nil,
-    pipeline = nil,
```
harrisoncramer, line 142: Why are we removing these settings? 

Additionally, if it's necessary to remove them they need to be added to our `health.lua` file under the `removed_settings_fields` section so that users know this functionality was deprecated.

lua/gitlab/utils/init.lua
```diff
@@ -485,10 +485,10 @@ end
 ---@param settings table|nil User defined popup settings
 ---@param width number? Override default width
 ---@param height number? Override default height
+---@param zindex number? Override default zindex
 ---@return table
 M.create_popup_state = function(title, settings, width, height, zindex)
-  local default_settings = require("gitlab.state").settings.popup
-  local user_settings = settings or {}
+  settings = M.merge(require("gitlab.state").settings.popup, settings or {})
```
harrisoncramer, line 491: Please do not variable shadow, either call the parameter something else or call this `merged_settings`

lua/gitlab/utils/popup.lua
```diff
@@ -0,0 +1,94 @@
+local u = require("gitlab.utils")
```
harrisoncramer, line 1: Let's just call this module `popup` and put it at the root level.

lua/gitlab/actions/comment.lua
```diff
@@ -229,27 +234,19 @@ M.create_comment_layout = function(opts)
     local text = u.get_buffer_text(M.comment_popup.bufnr)
     confirm_create_comment(text, range, unlinked, opts.discussion_id)
     vim.api.nvim_set_current_win(M.current_win)
-  end, miscellaneous.toggle_bool, miscellaneous.non_editable_popup_opts)
+  end, miscellaneous.toggle_bool, popup_utils.non_editable_popup_opts)
 
   ---Keybinding for focus on text section
   state.set_popup_keymaps(M.comment_popup, function(text)
```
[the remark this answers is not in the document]
harrisoncramer: Agreed, let's do that! And let's put popup at the root level, and move over `create_box_popup_state` and `create_popup_state` in there as well.

harrisoncramer:
This is a big improvement, thanks for making these changes. I've left some comments.

jakubbortlik:
> This is a big improvement, thanks for making these changes. I've left some comments.

Thanks for the review. I'll implement your suggestions and mark the PR for another review.

lua/gitlab/state.lua
jakubbortlik: My motivation was that the help and pipeline popups' configuration is limited - their height and width are calculated automatically and I thought it could be confusing to have the possibility to configure them but the configuration would not be applied.

But on second thoughts I think some users just still might want to configure opacity and border for these separately, so I'm adding the settings back in.

harrisoncramer (approved):

Issue #409, opened 2024-10-18
Show update status of discussion tree in winbar


jakubbortlik:
## Feature Description
It would be very handy to show the status of the discussion tree in the winbar so that the user can have an idea when it might be necessary to refresh the discussions.

I imagine that the winbar would contain something like `Updated: 8 minutes ago`, etc. and when an update is in progress, the `... ago` would be replaced by a simple animation. E.g., [neotest](https://github.com/nvim-neotest/neotest) by default cycles through these characters: `{ "/", "|", "\\", "-", "/", "|", "\\", "-" }` when showing the progress of a test. This would replace the info message `gitlab.nvim: Refreshing data...` which is not that helpful since the user does not know when the refreshing has finished, as the message does not disappear automatically.

Apart from that, I think that the `gitlab.utils.time_since` function could be changed so that the *minute* would be the smallest unit of time that is distinguished, anything under a minute would be just something like `Updated: moments ago`, unless the winbar would be updated every second, which I don't think is a good idea.

PR #432, opened 2024-11-23
Feat: Show update progress in winbar

jakubbortlik:
Hi @harrisoncramer, this is an attempt to implement #409.

With this PR, the winbar shows the time since the last update of the discussion data and it shows a simple animation when the data is being reloaded.

There are some shortcomings that would require some more work to solve:
1. The winbar has become quite crammed when using position `left` or `right` for the discussion tree, so that the "Inline Comments" section can easily move outside of the window. To fix this, we'd have to add some more elaborate winbar string calculation that would take the window width and the components into account. This can be done in a separate follow-up PR, I believe
2. The drafts data are loaded significantly faster* than regular discussion data, this results in a short inconsistency in the winbar when manipulating the drafts - the number of drafts is updated in the winbar faster than it is rendered in the discussion tree (and before the update spinner finishes), because rebuilding the discussion tree is postponed until after all discussion data are reloaded. I believe this could be fixed by changing the order in which the data are reloaded. Consider the following pseudocode.

The current order:
```lua
draft_notes.rebuild_view = function(unlinked, all)
    discussions.rebuild_view(unlinked, all)
end
```
would change to:
```lua
discussions.rebuild_view = function(unlinked, all)
    draft_notes.rebuild_view(unlinked, all)
end
```
I haven't tried this, but I believe it would make the discrepancy between the winbar and the discussion tree much smaller. I'm going to give this a try.

\* I've made some simple measurements, and it looks like even in a very small testing repo with a very small MR (13 items in the overview) it takes about 2 seconds for the discussion tree to reload. The drafts are reloaded in fractions of a second. I'm wondering if this is just how Gitlab works, or if there may be some way to make the go server do this faster.

### DIFF
--- a/lua/gitlab/utils/init.lua
+++ b/lua/gitlab/utils/init.lua
@@ -122,7 +122,7 @@ M.time_since = function(date_string, current_date_table)
   local time_diff = current_date - date
 
   if time_diff < 60 then
-    return "moments ago"
+    return "just now"
   elseif time_diff < 3600 then
     return M.pluralize(math.floor(time_diff / 60), "minute") .. " ago"
   elseif time_diff < 86400 then
--- a/lua/gitlab/utils/init.lua
+++ b/lua/gitlab/utils/init.lua
@@ -122,7 +122,7 @@ M.time_since = function(date_string, current_date_table)
   local time_diff = current_date - date
 
   if time_diff < 60 then
-    return M.pluralize(time_diff, "second") .. " ago"
+    return "moments ago"
   elseif time_diff < 3600 then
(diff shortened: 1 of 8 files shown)


lua/gitlab/utils/init.lua
```diff
@@ -122,7 +122,7 @@ M.time_since = function(date_string, current_date_table)
   local time_diff = current_date - date
 
   if time_diff < 60 then
-    return M.pluralize(time_diff, "second") .. " ago"
+    return "moments ago"
```
jakubbortlik, line 125: This corresponds to what Gitlab shows:
```suggestion
    return "just now"
```

PR #433, opened 2024-11-23
Fix: Syntax in discussion tree

jakubbortlik:
Hi @harrisoncramer, this is partly a follow-up PR to #432, since that PR changes the what `utils.time_since()` returns for times less than a minute, and this PR adds `just now` to the strings matched as `date` in the `gitlab` filetype syntax.

While I was touching the syntax file, I also fixed it so that the note header components (expanders, note creation time, (un)resolved symbols and the draft symbol) are only matched within the header, but not in the note body (before, even strings like "**5 minutes ago**ny" would be highlighted as time :)


harrisoncramer:
Hey @jakubbortlik I've merged in some changes to shorten the winbar, so this will need another look when you have a moment to resolve that conflict.

jakubbortlik:
> Hey @jakubbortlik I've merged in some changes to shorten the winbar, so this will need another look when you have a moment to resolve that conflict.

Hi Harrison, thanks for taking care of all the recent PRs! I really appreciate that you merged them in (or came up with alternatives like for the winbar shortening). This one should now also be ready for merge. The syntax highlighting is now only applied in the header of a node, unless somebody copies the text of the header and pastes it into the message :smile:

after/syntax/gitlab.vim
```diff
@@ -2,32 +2,26 @@ if filereadable($VIMRUNTIME . '/syntax/markdown.vim')
...
+let date = '\%(' . time_ago . '\|' . formatted_date . '\|just now\)'
+
+let published = date . ' \%(' . g:gitlab_discussion_tree_resolved . '\|' . g:gitlab_discussion_tree_unresolved . '\|' . g:gitlab_discussion_tree_unlinked . '\)\?'
+let state = ' \%(' . published . '\|' . g:gitlab_discussion_tree_draft . '\)'
+
+execute 'syntax match GitlabNoteHeader "' . expanders . username . state . '" contains=GitlabDate,GitlabUnresolved,GitlabUnlinked,GitlabResolved,GitlabExpander,GitlabDraft,GitlabUsername'
+
+execute 'syntax match GitlabDate "' . date . '" contained'
+execute 'syntax match GitlabUnresolved "' . g:gitlab_discussion_tree_unresolved . '" contained'
+execute 'syntax match GitlabUnlinked "' . g:gitlab_discussion_tree_unlinked . '" contained'
```
harrisoncramer, line 20: The unlinked highlights are not working on this branch in the discussion tree

jakubbortlik: In what sense does it not work? When I run neovim with the following config it works just fine, i.e., the `:Inspect` command prints `GitlabNoteHeader` and `GitlabUnlinked` :
```lua
-- This is a minimal init.lua that is used for debugging
local root = vim.fn.stdpath("run") .. "/nvim/mininit"
vim.fn.mkdir(root, "p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  {
    "jakubbortlik/gitlab.nvim",
    branch = "fix-syntax-in-discussion-tree",
    dependencies = {
      "MunifTanjim/nui.nvim",
      "nvim-lua/plenary.nvim",
      "sindrets/diffview.nvim",
      "stevearc/dressing.nvim",   -- Recommended but not required. Better UI for pickers.
      "nvim-tree/nvim-web-devicons" -- Recommended but not required. Icons in discussion tree.
    },
    enabled = true,
    build = function() require("gitlab.server").build(true) end, -- Builds the Go binary
    config = true,
  },
}

require("lazy").setup(plugins, {
  root = root .. "/plugins",
})
```

Highlighting does not work when I change the symbol for unlinked comments like this:
```lua
opts = {discussion_tree = {unlinked = "x"}},
```
But this is not a problem of this branch but of the way how the symbols are hardcoded in `lua/gitlab/colors.lua` lines 9-12. I'll push a fix for this in a moment.

If the problem still persists, please let me know what it looks like how I can reproduce it.

Issue #424, opened 2024-11-18
Checking for changes before making comments is incorrect


jakubbortlik:
## Prerequsities

- [x] I'm on the latest version of the plugin
- [x] I've installed the required dependencies
- [x] I've run `:h gitlab.nvim.troubleshooting` and followed the steps there

## Bug Description
The function `has_clean_tree` in `lua/gitlab/git.lua` line 30 returns a string and hence evaluates to `true` if the worktree is dirty. This causes some problems, e.g., the `actions.comment.create_comment` function in `lua/gitlab/actions/comment.lua` line 258 seems to check for changed files and to refuse creating the comment in case of a dirty tree, but fails because this `if` statement evaluates to `false` even if `imply_local` is true and there are modified tracked files (edit: `not ""` evaluates to `false` just like `not "M modified_file.txt"`): 
```lua
  if state.settings.reviewer_settings.diffview.imply_local and (is_modified or not has_clean_tree) then
```

It could be fixed by changing the line to this (`has_clean_tree` is actually used like this e.g. in `lua/gitlab/actions/merge_requests.lua` line 18):

```lua
  if state.settings.reviewer_settings.diffview.imply_local and (is_modified or has_clean_tree ~= "") then
```

But it would be more appropriate to fix the function and all the places where it is used:
```diff
diff --git a/lua/gitlab/git.lua b/lua/gitlab/git.lua
index aa74abc..e415eba 100644
--- a/lua/gitlab/git.lua
+++ b/lua/gitlab/git.lua
@@ -26,9 +26,10 @@ M.branches = function(args)
 end
 
 ---Checks whether the tree has any changes that haven't been pushed to the remote
----@return string|nil, string|nil
+---@return boolean, string|nil
 M.has_clean_tree = function()
-  return run_system({ "git", "status", "--short", "--untracked-files=no" })
+  local changes, err = run_system({ "git", "status", "--short", "--untracked-files=no" })
+  return changes == "", err
 end
 
 ---Gets the base directory of the current project
```
## Steps to reproduce
1. Check out the feature branch
2. Open Neovim
3. Open a tracked file in a modifiable buffer
4. Change the file and save
5. Go to the reviewer and run `lua require("gitlab").create_comment()`

This should raise an error, but instead, the comment popup is opened.

The fixed function should then be applied to also prevent `create_multiline_comment` and `create_comment_suggestion`, ideally in a unified logic that would be applied for all the comment creating functions.

PR #434, opened 2024-11-26
Fix: Checking whether comment can be created

jakubbortlik:
Hi @harrisoncramer, in this PR I address #424, by making the following changes:

1. All checks whether a comment can be created are now made for all three functions: `create_comment`, `create_comment_suggestion` and `create_multiline_comment` in a dedicated function `gitlab.actions.comment.can_make_comment()`.
2. The checks are made in a more meaningful order (e.g., checking that cursor is in the reviewer is made before the check that current mode is visual mode for multiline comments).
3. The check for clean working tree before making comments is now correct and it's made on a per-file basis, i.e., unsaved/uncommitted changes in a different file do not block a comment in a good file.
4. Similarly, with a dirty working tree, choosing a MR is not blocked if that would not lead to switching the branch.
5. User-specified `imply_local = true` is overridden in the settings if the working tree is dirty. Before, it was possible that the setting `imply_local` was true, while the option was not actually applied to the `DiffviewOpen` command and so commenting was unnecessarily blocked for modified files.

There are some other improvements:
1. Visual mode is exited when multiline comment creation fails.
2. The ERROR level is used for all messages when a comment cannot be created.
3. A draft reply can now be created even if the discussion tree is "detached" from the reviewer.

I'll be grateful if you find the time to review and merge this.

### DIFF
--- a/lua/gitlab/actions/comment.lua
+++ b/lua/gitlab/actions/comment.lua
@@ -236,6 +236,7 @@ end
 --- on the changed/updated line in the current MR
 M.create_multiline_comment = function()
   if not M.can_create_comment(true) then
+    u.press_escape()
     return
   end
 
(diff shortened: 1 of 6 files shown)


harrisoncramer:
This looks great, thank you for the contribution. The refactor and consolidation of the comment checks was needed.

PR #435, opened 2024-11-26
Feat: Adapt winbar string to window width

jakubbortlik:
This is a followup to #432 that makes sure that if the discussion tree is opened in a right split, the `winbar` is not just blindly truncated from the left which would hide useful information (number of comments), but individual components of the `winbar` string are dynamically shortened to fit in the window.

### DIFF
--- a/lua/gitlab/actions/discussions/winbar.lua
+++ b/lua/gitlab/actions/discussions/winbar.lua
@@ -143,6 +143,68 @@ local add_drafts_and_resolvable = function(
   return base_title
 end
 
+-- Returns true if the input string (after removing highlighting and alignment strings) is wider
+-- than the discussion split.
+---@return boolean
+local too_long = function(str)
+  local d = require("gitlab.actions.discussions")
+  return #str:gsub("%%#[^#]+#", ""):gsub("%%= ", "") > vim.fn.winwidth(d.split.winid)
+end
+
+-- Returns the input winbar string shortened to fit into the windo width.
+---@return string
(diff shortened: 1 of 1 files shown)


lua/gitlab/actions/discussions/winbar.lua
```diff
@@ -143,6 +143,60 @@ local add_drafts_and_resolvable = function(
...
+---@return boolean
+local too_long = function(str)
+  local d = require("gitlab.actions.discussions")
+  return #str:gsub("%%#[^#]+#", "") > vim.fn.winwidth(d.split.winid)
+end
+
+-- Returns the input winbar string shortened to fit into the windo width.
+---@return string
+local adapt_to_winwidth = function(str)
+  if too_long(str) then
```
jakubbortlik, line 157: The solution is not particularly elegant, but unfortunately, Lua tables do not keep order of insertion, so it's not possible to do simply something like this if we want the `winbar` to adapt in a predictable way:
```lua
local shortenings = {
  ["Inline "] = "".
  [" Mode"] = "",
}
for pattern, replacement in pairs(shortenings) do
  if too_long(str) then
    str:gsub(pattern, replacement)
  end
end
```

harrisoncramer:
Hey Jakub, I actually shortened the winbar quite a bit, so I'm not sure this is still relevant -- if so, you may have to start anew from latest develop.

harrisoncramer:
In fact, you're probably just better off creating a new MR if you want to make changes to the winbar, I'm going to close this one out.

PR #436, opened 2024-12-02
fix: json settings not readable on Windows

pcorbineau:
backslash doublequote created by `vim.json.encode`. This help Windows command parsing

Am I the first user of the plugin on Windows? Because, as it stands, the server can't be launched because of the parsing of the json settings into command parameters.


harrisoncramer:
Interesting, you're not and I'm surprised this hasn't come up before. Does this resolve the issue for you and make the plugin functional?

pcorbineau:
I updated tracking base branch to develop.
It could be cool to have someone else who can reproduce the bug on main on Windows and let him try this branch before merge.

harrisoncramer:
Thank you for this. Could you please run Stylua locally and recommit this change? Your file is failing our linting check.

Issue #347, opened 2024-08-15
Default suggestion keymaps blocks LazyVim flash searching


briandipalma:
## Prerequsities

- [x] I'm on the latest version of the plugin
- [x] I've installed the required dependencies
- [x] I've run `:h gitlab.nvim.troubleshooting` and followed the steps there

## Setup Configuration and Environment

I'm trying to use the default config, so I deleted my keymaps.

## Bug Description

Not a bug, as you can change the keymaps but it's very common for `s` to be the flash.nvim trigger key which means the suggestion keymap breaks your ability to move around diffs and sometimes even opened files which is very jarring (I think there's a bug in diffview where buffer local settings can leak into opened files but I don't know how it's triggered). Although I'm not totally sure what keymap to suggest as an alternative to be honest I think it would be nice if it didn't clash. Maybe just use `cs` only and drop the top level  `s` keymap?

harrisoncramer:
Hey Brian, thanks for filing the feedback. We _just_ added these default keymaps and this didn't come up during that MR review... I'm not inclined to change them again. The easiest fix here is just to change the mapping to something else if you use that particular plugin!

jakubbortlik:
Hi, as for "the suggestion keymap breaks your ability to move around diffs and *sometimes even opened file*" - this may be caused by the `del_reviewer_keymaps` not to be called properly within the [callback for leaving the reviewer](https://github.com/harrisoncramer/gitlab.nvim/blob/aa5d3c1f523c969a4f381c57d00c9a206a6dc7e3/lua/gitlab/actions/discussions/init.lua#L86). It would help debug if you @brian could find a way to reproduce this.

Vim does not allpw setting a window-local keymap so that if you set a *buffer*-local mapping it is available in that buffer in any window/tab. I thought I fixed this by adding that callback that deletes the keymaps when leaving the reviewer, but appafently the autocommand does not fire always...

briandipalma:
I've never really understood why this happens when using diffview but I'm used to it by now, see this issue https://github.com/sindrets/diffview.nvim/issues/330 which leaves me with absolute line numbers instead of relative. You appear to be relying on diffview so that explains why deleting the keymaps is randomly not triggering when you open a file from a diff. I'll try and see if I can recreate it consistently but I don't consider this a bug with gitlab.nvim. I think it's probably a diffview or Neovim bug so I'd be fine with closing this.

PR #446, opened 2024-12-11
Fix: Restore buffer local settings outside reviewer

jakubbortlik:
This PR fixes an issue originally reported in #347. After a closer inspection it *is* a gitlab.nvim bug, as currently we are removing keymaps and resetting 'modifiable' only for the currently active buffer, but not for other buffers previously shown by Diffview.

This PR fixes the issue by setting and deleting keymaps when entering any window that contains a Diffview buffer or when switching to a Diffview buffer without changing the window (i.e., cursor is already in the reviewer).

Edit: Apart from this fix, this PR also limits refreshing diagnostics to when the `reviewer` is entered, as opposed to the previous way when diagnostics were refreshed every time the reviewer buffer was changed. This speeds up the file switching significantly (although this is practically only noticeable when *holding down* `<tab>` to cycle between the diffs quickly which I don't think is something people do frequently :smile:).


jakubbortlik:
@briandipalma in August you reported the [issue](https://github.com/harrisoncramer/gitlab.nvim/issues/347#issuecomment-2294995072) that I'm fixing in this PR, could you please try out [this branch](https://github.com/jakubbortlik/gitlab.nvim/tree/fix-buffer-local-settings-outside-reviewer) and let us know if it fixes the issue?

lua/gitlab/reviewer/init.lua
```diff
@@ -426,23 +446,33 @@ local del_keymaps = function(bufnr, keymaps)
...
 end
 
---- Deletes keymaps from both buffers in the reviewer.
-M.del_reviewer_keymaps = function()
-  -- Require keymaps only after user settings have been merged with defaults
-  local keymaps = require("gitlab.state").settings.keymaps
-  if keymaps.disable_all or keymaps.reviewer.disable_all then
-    return
-  end
+--- Set up autocaommands that will take care of setting and unsetting buffer-local options and keymaps
```
jakubbortlik, line 449:
```suggestion
--- Set up autocommands that will take care of setting and unsetting buffer-local options and keymaps
```

briandipalma:
I've just gone on holidays so I doubt I'll get the time to test this until after Christmas, I'll take a look in January.

briandipalma:
I've tried to recreate the issue but I'm failing so maybe it's gone, it was random so who knows. What I'll do is keep an eye out for the issue and if I see it I'll try this branch out. Sorry about taking so long.

jakubbortlik:
> I've tried to recreate the issue but I'm failing so maybe it's gone, it was random so who knows. What I'll do is keep an eye out for the issue and if I see it I'll try this branch out. Sorry about taking so long.

Hi Brian, thanks for having a look at this. Here's how you can reproduce the bug:
1. Open a MR with multiple modified files
2. In the Diffview File Panel press `<tab>` several times (or `lua require("diffview.actions").select_next_entry()`)
3. run `:ls<cr>`, you should see in the buffer list, that the previously visited buffers are `nomodifiable`
4. In the File Panel or in one of the reviewer windows press `gf` which should open the file in another tab, where the current buffer is now `modifiable` and the commenting keymaps are removed (`:map c` should not show something like `v  c           *@<Lua 1782: ~/projects/gitlab.nvim/lua/gitlab/reviewer/init.lua:384>\n                 Create comment for selected text`.
5. In the same tab (i.e., not the `gitlab.nvim` tab) open one of the other reviewed files (e.g., by running `:bprev`). This file will be `nomodifiable` and `:map c` will show that the commenting mappings are still there

In the `jakubbortlik:fix-buffer-local-settings-outside-reviewer` branch, this should be fixed, i.g., step 3 will still show that the buffer you've just cycled through with Diffview's `<tab>` keymap are still shown as `nomodifiable`, but when you do step 5, you'll see that upon *visiting* the buffer in another tab, the options and mappings are fixed.

PR #448, opened 2024-12-26
Draft: Fix: Go Gitlab Project Has Moved

### LINKED ISSUE #447: New Repository for Go Gitlab
Go Gitlab is moving to Gitlab and we may want to update this dependency: https://gitlab.com/gitlab-org/api/client-go

harrisoncramer:
This required changing Go to 1.23, unless I'm mistaken, which will make this an unfortunate breaking change for downstream users. To fix, users will have to update their local Go versions to 1.23.4, as the plugin requires this local version of Go in order to compile. 

It's worth looking at shipping precompiled versions of the plugins moving forward to avoid this for end-users in the future.

Fixes #447. 

Note: This is a #MAJOR breaking change, that will require users to update their versions of Go locally to latest (1.23.4)


camfowler:
I was receiving 500 errors when trying to add comments to an MR, and changing to this branch fixed it.

lua/gitlab/health.lua
```diff
@@ -6,11 +6,11 @@ local M = {}
 local function check_go_version()
   local go_version = io.popen("go version"):read("*a")
   if go_version then
-    local major, minor, _ = go_version:match("(%d+)%.(%d+)%.(%d+)")
-    if major and tonumber(major) >= 1 and tonumber(minor) >= 19 then
+    local major, minor, patch = go_version:match("(%d+)%.(%d+)%.(%d+)")
+    if major and tonumber(major) >= 1 and tonumber(minor) >= 23 and tonumber(patch) >= 4 then
```
camfowler, line 10: My go version is `go version go1.24.0 darwin/arm64`
Which fails this logic check, because whilst 1 and 24 are ok 0 >= 4 fails, even though 1.24.0 is > 1.23.4

harrisoncramer: Thank you, I'll update fix in a separate PR.

jakubbortlik:
I've been testing `client-go` because of other features and it turns out, that there will be some more work to be done when switching from `xanzy/go-gitlab`, e.g., the [ListProjectMergeRequests](https://gitlab.com/gitlab-org/api/client-go/-/commit/42ec248d8b3bd359b57740e4f9b0f3d03ba02b97#d3c8d5bf70b5b7a53d739f88dd276ea07e77532b_230_244) now returns `BasicMergeRequests`, rather than `MergeRequests`.

harrisoncramer:
I'll work on this migration shortly. We'll probably have to have a migration tool such that when someone upgrades, it prompts them to install the latest version of Go or something. Will think on it.

PR #449, opened 2024-12-26
Fix: Remove Retry Logic

harrisoncramer:
This MR changes the configuration to the HTTP client so that it doesn't retry failed attempts. This behavior was causing problems when Gitlab fails with a 500 status code, but actually publishes something.

### DIFF
--- a/cmd/app/client.go
+++ b/cmd/app/client.go
@@ -68,6 +68,7 @@ func NewClient() (*Client, error) {
 
 	retryClient := retryablehttp.NewClient()
 	retryClient.HTTPClient.Transport = tr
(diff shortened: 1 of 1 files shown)


harrisoncramer:
@jakubbortlik I think this will prevent posting multiple notes/comments, this HTTP client has retry logic that I believe is being triggered by general HTTP errors from Gitlab: https://github.com/hashicorp/go-retryablehttp

jakubbortlik:
Nice! I haven't really tried to find a way to reproduce the issue, but I just hope it will not reappear after this fix. Thanks!

PR #467, opened 2025-01-29
Feat: Add MR URL to the summary details

jakubbortlik:
This is a small PR that adds the MR's Web URL to the Details of the summary popup. I sometimes find it useful to be able to *see* the URL. A workaround is to use `glu` to copy the url to clipboard and then it is shown in a message, but sometimes I just don't want to pollute the clipboard just to see a piece of information.

### DIFF
--- a/doc/gitlab.nvim.txt
+++ b/doc/gitlab.nvim.txt
@@ -302,6 +302,7 @@ you call this function with no values the defaults will be used:
           "delete_branch",
           "squash",
           "labels",
+          "web_url",
         },
       },
       discussion_signs = {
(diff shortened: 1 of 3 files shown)


jakubbortlik:
The [check](https://github.com/harrisoncramer/gitlab.nvim/actions/runs/13041578430/job/36384436077?pr=467) is failing because Neovim 0.10.4 does not have the asset `nvim-linux64.tar.gz`, but it has `nvim-linux-x86_64.tar.gz`.

There's already a [PR](https://github.com/rhysd/action-setup-vim/pull/40) to fix this in the `action-setup-vim` repository.

harrisoncramer:
Will approve and merge this once the pipeline is fixed, thanks!

PR #470, opened 2025-01-31
Small Bug Fixes

harrisoncramer:


jakubbortlik:
#446 included the idea to make cycling the reviewed files faster by only updating diagnostics upon entering the files for the first time. This is a brilliant idea, except it doesn't work. Diagnostics are now not displayed when cycling over the files repeatedly.

A simple fix is to add back the refreshing to the `set_callback_for_file_changed` callback:
```diff
diff --git a/lua/gitlab/actions/discussions/init.lua b/lua/gitlab/actions/discussions/init.lua
index 3a32d4e..5b93647 100644
--- a/lua/gitlab/actions/discussions/init.lua
+++ b/lua/gitlab/actions/discussions/init.lua
@@ -74,6 +74,7 @@ M.initialize_discussions = function()
   state.discussion_tree.last_updated = os.time()
   signs.setup_signs()
   reviewer.set_callback_for_file_changed(function(args)
+    M.refresh_diagnostics()
     reviewer.update_winid_for_buffer(args.buf)
   end)
   reviewer.set_callback_for_reviewer_enter(function()
```
However, that will make the reviewer slow again, because every time the diagnostics are refreshed, the complete list of discussions is filtered several times. I'm working on a fix that will consist of preparing the diagnostics once and then selecting the right diagnostics for the relevant buffer. It will take some time, so if you @harrisoncramer are planning to release #470 soon, I suggest to include the above fix.

PR #473, opened 2025-02-03
fix: show diagnostics when cycling buffers

jakubbortlik:
Fixes the bug introduced by #446 which causes the diagnostics not to be shown properly.


jakubbortlik:
Superseded by #474.

PR #474, opened 2025-02-04
fix: make cycling reviewed files faster

jakubbortlik:
Addresses the issue described in [this comment](https://github.com/harrisoncramer/gitlab.nvim/pull/470#issuecomment-2632298958).

This fixes the diagnostics placement while also making cycling Diffview files faster by pre-computing diagnostics and removing unnecessary list filtering.

### DIFF
--- a/doc/gitlab.nvim.txt
+++ b/doc/gitlab.nvim.txt
@@ -316,6 +316,7 @@ you call this function with no values the defaults will be used:
           comment = "→|",
           range = " |",
         },
+        skip_old_revision_discussion = false, -- Don't show diagnostics for discussions that were created for earlier MR revisions
       },
       pipeline = {
         created = "",
--- a/lua/gitlab/actions/common.lua
+++ b/lua/gitlab/actions/common.lua
@@ -265,7 +265,7 @@ M.get_line_number_from_node = function(root_node)
 end
 
 -- This function (settings.keymaps.discussion_tree.jump_to_reviewer) will jump the cursor to the reviewer's location associated with the note. The implementation depends on the reviewer
-M.jump_to_reviewer = function(tree, callback)
+M.jump_to_reviewer = function(tree)
   local node = tree:get_node()
   local root_node = M.get_root_node(tree, node)
   if root_node == nil then
@@ -278,7 +278,6 @@ M.jump_to_reviewer = function(tree, callback)
     return
   end
   reviewer.jump(root_node.file_name, line_number, is_new_sha)
-  callback()
 end
 
 -- This function (settings.keymaps.discussion_tree.jump_to_file) will jump to the file changed in a new tab
--- a/lua/gitlab/actions/discussions/init.lua
+++ b/lua/gitlab/actions/discussions/init.lua
@@ -744,6 +744,16 @@ M.set_tree_keymaps = function(tree, bufnr, unlinked)
     })
   end
 
+  if keymaps.discussion_tree.print_node then
+    vim.keymap.set("n", keymaps.discussion_tree.print_node, function()
+      common.print_node(tree)
+    end, {
+      buffer = bufnr,
+      desc = "Print current node (for debugging)",
+      nowait = keymaps.discussion_tree.print_node_nowait,
+    })
+  end
+
   if keymaps.discussion_tree.add_emoji then
     vim.keymap.set("n", keymaps.discussion_tree.add_emoji, function()
       M.add_emoji_to_note(tree, unlinked)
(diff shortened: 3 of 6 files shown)


lua/gitlab/actions/discussions/init.lua
```diff
@@ -74,6 +74,7 @@ M.initialize_discussions = function()
   state.discussion_tree.last_updated = os.time()
   signs.setup_signs()
   reviewer.set_callback_for_file_changed(function(args)
+    diagnostics.place_diagnostics(args.buf)
```
jakubbortlik, line 77: Placing diagnostics with the `bufnr` parameter makes sure that diagnostics are not places twice - the callback created with `reviewer.set_callback_for_file_changed` fires for each reviewer buffer (OLD and NEW) separately, so with just `M.refresh_diagnostics()` in place of `diagnostics.place_diagnostics(args.buf)`, diagnostics would be placed in OLD and NEW buff twice.

lua/gitlab/actions/common.lua
```diff
@@ -265,7 +265,7 @@ M.get_line_number_from_node = function(root_node)
 end
 
 -- This function (settings.keymaps.discussion_tree.jump_to_reviewer) will jump the cursor to the reviewer's location associated with the note. The implementation depends on the reviewer
-M.jump_to_reviewer = function(tree, callback)
```
jakubbortlik, line 268: The callback is not necessary anymore since it's called automatically when the respective buffer enters the reviewer window.

doc/gitlab.nvim.txt
```diff
@@ -316,6 +316,7 @@ you call this function with no values the defaults will be used:
           comment = "→|",
           range = " |",
         },
+        skip_old_revision_discussion = false, -- Don't show diagnostics for discussions that were created for earlier MR revisions
```
jakubbortlik, line 319: This was undocumented but still possible to change.

lua/gitlab/actions/discussions/init.lua
```diff
@@ -743,6 +744,16 @@ M.set_tree_keymaps = function(tree, bufnr, unlinked)
     })
   end
 
+  if keymaps.discussion_tree.print_node then
```
jakubbortlik, line 747: The mapping was still in the `lua/gitlab/state.lua` and in the docs, so I assume this `keymap.set` call was removed by accident. In any case, I've been missing the possibility to see node content quite a lot when working on several issues recently, so I'd appreciate if we can get it back.

lua/gitlab/indicators/diagnostics.lua
```diff
@@ -85,47 +85,64 @@ local create_multiline_diagnostic = function(d_or_n)
   }, d_or_n)
 end
 
----Set diagnostics in currently new SHA.
+---Set diagnostics in the given buffer.
 ---@param namespace number namespace for diagnostics
+---@param bufnr number the bufnr for placing the diagnostics
 ---@param diagnostics table see :h vim.diagnostic.set
 ---@param opts table? see :h vim.diagnostic.set
-local set_diagnostics_in_new_sha = function(namespace, diagnostics, opts)
```
jakubbortlik, line 92: By providing the `bufnr` we no longer need two separate functions for the `new_sha` and the `old_sha`.

lua/gitlab/indicators/diagnostics.lua
```diff
@@ -85,47 +85,64 @@ local create_multiline_diagnostic = function(d_or_n)
...
-    require("gitlab.indicators.signs").clear_signs()
-    M.clear_diagnostics()
-    local filtered_discussions = indicators_common.filter_placeable_discussions()
-    if filtered_discussions == nil then
+    local file_discussions = List.new(M.placeable_discussions):filter(function(discussion_or_note)
+      local note = discussion_or_note.notes and discussion_or_note.notes[1] or discussion_or_note
+      -- Surprisingly, the following line works even if I'd expect it should match new/old paths
+      -- with a/b buffers the other way round!
+      return note.position.new_path == view.cur_layout.a.file.path
+        or note.position.old_path == view.cur_layout.b.file.path
```
jakubbortlik, line 132: By "the following line works" I mean that it helps to place more diagnostics in the reviewer than when matching `new_path` with the `b` file, e.g., in a testing MR, I've renamed/moved `file2.txt` to `tmp/file2.txt` and when I debug-print the variables on this line, I get the following values:
```
note.position.new_path="file2.txt"
note.position.old_path="tmp/file2.txt"
view.cur_layout.b.file.path="tmp/file2.txt"
view.cur_layout.a.file.path="file2.txt"
```
The surprisingly, the path to which the file has been renamed is stored as `note.position.old_path`.

PR #476, opened 2025-02-18
Fix: Store reviewer data before creating comment popup

### LINKED ISSUE #475: Changing reviewed file while creating a comment places comment on wrong file
## Prerequsities

## Bug Description

When creating a comment in the reviewer, it is sometimes useful to get back from the comment popup to the reviewer, e.g., to scroll to see the context (sometimes a user can just by accident leave the popup and end up in the reviewer).

When this happens and the user happens to cycle the reviewer files (e.g., by pressing the default Diffview `<tab>` keybinding) and then comes back to the comment popup and finishes the comment, it will be created for the last active file and not for the one for which it was originally created.

This is because the reviewer data are only retrieved in the [confirm_create_comment](https://github.com/harrisoncramer/gitlab.nvim/blob/a893f61bfc496e2468bdb1f9cc674e1b5c2e542b/lua/gitlab/actions/comment.lua#L78) callback and not right when starting the comment.

## Reproduction Steps

1. Check out the feature branch
2. Open Neovim
3. Start a review with at least 2 files that contain some text
4. Press `ss` or `cc` on file A
5. Within the comment popup press `<c-w><c-w>` to move to the next window (or navigate to the Diffview File Panel in any other way)
6. Cycle the reviewed files to file B
7. Press `<c-w>W` to get back to the comment popup
8. Press `ZZ` to apply the comment
9. Using the `by_file_name` discussion tree listing observe that the comment is created for file B instead of A

## Suggested solution

We move the call to `get_reviewer_data` to the [create_comment_layout](https://github.com/harrisoncramer/gitlab.nvim/blob/a893f61bfc496e2468bdb1f9cc674e1b5c2e542b/lua/gitlab/actions/comment.lua#L161) function, build the [location](https://github.com/harrisoncramer/gitlab.nvim/blob/a893f61bfc496e2468bdb1f9cc674e1b5c2e542b/lua/gitlab/actions/comment.lua#L84) there and pass the result to `confirm_create_comment` as an argument (instead of just passing [visual_range](https://github.com/harrisoncramer/gitlab.nvim/blob/a893f61bfc496e2468bdb1f9cc674e1b5c2e542b/lua/gitlab/actions/comment.lua#L28)).

## Suggested improvement
When creating a comment, we place signs into the reviewer's sign column that will mark the selected range until the comment is applied or aborted.

jakubbortlik:
Resolves #475.

Hi @harrisoncramer, I'd be grateful if you could review this PR. It resolves #475 by doing the following:
1. Reviewer data are collected *before creating* the comment popup, not just *before confirming* the comment.
2. The reviewer data are now managed *inside* of the `Location` class.

I'll add some more comments to the changed code itself.

### DIFF
--- a/lua/gitlab/annotations.lua
+++ b/lua/gitlab/annotations.lua
@@ -125,6 +125,8 @@
 ---@field opposite_bufnr integer
 ---@field new_line_from_buf integer
 ---@field old_line_from_buf integer
+---@field new_sha_focused boolean
+---@field current_win_id integer
 
 ---@class LocationData
 ---@field old_line integer | nil
--- a/lua/gitlab/annotations.lua
+++ b/lua/gitlab/annotations.lua
@@ -120,8 +120,6 @@
 ---Relevant for renamed files only, the name of the file in the previous commit
 ---@field old_file_name string
 ---@field current_bufnr integer
----@field new_sha_win_id integer
----@field old_sha_win_id integer
 ---@field opposite_bufnr integer
 ---@field new_line_from_buf integer
 ---@field old_line_from_buf integer
--- a/lua/gitlab/hunks.lua
+++ b/lua/gitlab/hunks.lua
@@ -230,9 +230,9 @@ end
 ---This is in order to build the payload for Gitlab correctly by setting the old line and new line.
 ---@param old_line number|nil
 ---@param new_line number|nil
----@param is_current_sha_focused boolean
+---@param new_sha_focused boolean
 ---@return string|nil
-function M.get_modification_type(old_line, new_line, is_current_sha_focused)
+function M.get_modification_type(old_line, new_line, new_sha_focused)
   local hunk_and_diff_data = parse_hunks_and_diff(state.INFO.diff_refs.base_sha)
   if hunk_and_diff_data.hunks == nil then
     return
@@ -240,7 +240,7 @@ function M.get_modification_type(old_line, new_line, is_current_sha_focused)
 
   local hunks = hunk_and_diff_data.hunks
   local all_diff_output = hunk_and_diff_data.all_diff_output
-  return is_current_sha_focused and get_modification_type_from_new_sha(new_line, hunks, all_diff_output)
+  return new_sha_focused and get_modification_type_from_new_sha(new_line, hunks, all_diff_output)
     or get_modification_type_from_old_sha(old_line, new_line, hunks, all_diff_output)
 end
 
--- a/lua/gitlab/indicators/common.lua
+++ b/lua/gitlab/indicators/common.lua
@@ -12,7 +12,6 @@ local M = {}
 
 ---Return true if discussion has a placeable diagnostic, false otherwise.
 ---@param note NoteWithValues
----@param file string
 ---@return boolean
 local filter_discussions_and_notes = function(note)
   ---Do not include unlinked notes
--- a/lua/gitlab/reviewer/init.lua
+++ b/lua/gitlab/reviewer/init.lua
@@ -153,8 +153,9 @@ end
 
 ---Get the data from diffview, such as line information and file name. May be used by
 ---other modules such as the comment module to create line codes or set diagnostics
+---@param current_win integer The ID of the currently focused window
 ---@return DiffviewInfo | nil
-M.get_reviewer_data = function()
+M.get_reviewer_data = function(current_win)
   local view = diffview_lib.get_current_view()
   if view == nil then
     return
@@ -177,7 +178,7 @@ M.get_reviewer_data = function()
   local new_line = vim.api.nvim_win_get_cursor(new_win)[1]
   local old_line = vim.api.nvim_win_get_cursor(old_win)[1]
 
-  local new_sha_focused = M.is_new_sha_focused()
+  local new_sha_focused = M.is_new_sha_focused(current_win)
 
   local modification_type = hunks.get_modification_type(old_line, new_line, new_sha_focused)
   if modification_type == nil then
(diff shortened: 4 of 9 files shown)


lua/gitlab/annotations.lua
```diff
@@ -120,11 +120,11 @@
 ---Relevant for renamed files only, the name of the file in the previous commit
 ---@field old_file_name string
 ---@field current_bufnr integer
----@field new_sha_win_id integer
----@field old_sha_win_id integer
```
jakubbortlik, line 124: These two fields were only ever used for figuring out whether the new SHA is focused. Instead, we now just store this information directly as `new_sha_focused` below.

lua/gitlab/hunks.lua
```diff
@@ -230,17 +230,17 @@ end
 ---This is in order to build the payload for Gitlab correctly by setting the old line and new line.
 ---@param old_line number|nil
 ---@param new_line number|nil
----@param is_current_sha_focused boolean
+---@param new_sha_focused boolean
```
jakubbortlik, line 233: I find `current` here a little confusing, in other places it's used for `currently` selected window (which could be the `old SHA`).

lua/gitlab/indicators/common.lua
```diff
@@ -12,7 +12,6 @@ local M = {}
 
 ---Return true if discussion has a placeable diagnostic, false otherwise.
 ---@param note NoteWithValues
----@param file string
```
jakubbortlik, line 15: Leftover from previous PR.

lua/gitlab/reviewer/init.lua
```diff
@@ -186,32 +190,32 @@ M.get_reviewer_data = function()
...
 ---Return whether user is focused on the new version of the file
+---@param current_win integer The ID of the currently focused window
 ---@return boolean
-M.is_current_sha_focused = function()
+M.is_new_sha_focused = function(current_win)
   local view = diffview_lib.get_current_view()
   local layout = view.cur_layout
   local b_win = u.get_window_id_by_buffer_id(layout.b.file.bufnr)
   local a_win = u.get_window_id_by_buffer_id(layout.a.file.bufnr)
-  local current_win = require("gitlab.actions.comment").current_win
```
jakubbortlik, line 214: I removed `current_win` from the `gitlab.actions.comment` module as I find it clearer to pass it as an explicit parameter.

lua/gitlab/reviewer/location.lua
```diff
@@ -19,16 +20,21 @@ local state = require("gitlab.state")
 
 local Location = {}
 Location.__index = Location
----@param reviewer_data DiffviewInfo
----@param visual_range LineRange | nil
```
jakubbortlik, line 23: I find it clearer when the reviewer data is managed inside the `Location` class itself rather than passing it as a parameter - this makes the use of `Location` much simpler than it was in `lua/gitlab/actions/comment.lua` on line 78 and forward.

lua/gitlab/reviewer/location.lua
```diff
@@ -19,16 +20,21 @@ local state = require("gitlab.state")
...
+  if reviewer_data == nil then
+    return nil
+  end
   local location = {}
   local instance = setmetatable(location, Location)
   instance.reviewer_data = reviewer_data
-  instance.visual_range = visual_range
   instance.base_sha = state.INFO.diff_refs.base_sha
   instance.head_sha = state.INFO.diff_refs.head_sha
+  instance:build_location_data()
```
jakubbortlik, line 37: Building location data in the constructor makes usage of the class a little simpler (fewer needs for `nil` checks).

lua/gitlab/reviewer/location.lua
```diff
@@ -51,30 +59,29 @@ function Location:build_location_data()
...
-    location_data.new_line = reviewer_data.new_line_from_buf
+    self.location_data.old_line = reviewer_data.old_line_from_buf
+    self.location_data.new_line = reviewer_data.new_line_from_buf
   end
 
-  self.location_data = location_data
-  if visual_range == nil then
-    return
-  else
+  if end_line > start_line then
```
jakubbortlik, line 74: We now *always* store the `visual_range`, so we can't check if it's `nil`. Before, it was `nil` when we explicitly passed the `ranged` parameter to `create_comment_layout` as `true`, now we don't need that parameter anymore as it's trivial to figure out automatically whether a comment is ranged or not.

lua/gitlab/reviewer/location.lua
```diff
@@ -90,9 +97,7 @@ end
 ---@param line number
 ---@return number|nil
 function Location:get_line_number_from_new_sha(line)
-  local reviewer = require("gitlab.reviewer")
-  local is_current_sha_focused = reviewer.is_current_sha_focused()
-  if is_current_sha_focused then
+  if self.reviewer_data.new_sha_focused then
```
jakubbortlik, line 100: Saving the `new_sha_focused` once when building the `Location` enables us to get rid of quite a few `is_current_sha_focused()` calls.

lua/gitlab/reviewer/location.lua
```diff
@@ -131,32 +134,24 @@ end
...
 
-  local current_line = vim.api.nvim_win_get_cursor(win_id)[1]
+  local current_line = vim.api.nvim_win_get_cursor(self.reviewer_data.current_win_id)[1]
   return current_line
 end
 
--- Given a new_line and old_line from the start of a ranged comment, returns the start
--- range information for the Gitlab payload
----@param visual_range LineRange
----@return ReviewerLineInfo|nil
```
jakubbortlik, line 148: The method wasn't really returning anything.

lua/gitlab/utils/init.lua
```diff
@@ -587,19 +587,20 @@ end
...
 ---@return integer start,integer end Start line and end line
 M.get_visual_selection_boundaries = function()
-  M.press_escape()
-  local start_line = vim.api.nvim_buf_get_mark(0, "<")[1]
-  local end_line = vim.api.nvim_buf_get_mark(0, ">")[1]
+  local start_line = vim.fn.line("v")
+  local end_line = vim.fn.line(".")
+  if start_line > end_line then
+    start_line, end_line = end_line, start_line
+  end
```
jakubbortlik, line 603: This method allows us to get rid of the `M.press_escape()` call.

Issue #486, opened 2025-03-20
Adding comment returns 500 Internal Server error


nickyhuyskens:
## Prerequsities

- [ x] I'm on the latest version of the plugin - yes, both the default and develop branch give this issue
- [ x] I've installed the required dependencies
- [ x] I've run `:h gitlab.nvim.troubleshooting` and followed the steps there

## Setup Configuration and Environment

Default configuration with url and token set as environment variables

## Bug Description

Whenever I add a comment on a diff (cc followed by typing the comment and then ZZ), I get an error below with the message: gitlab.nvim: Could not create discussion: POST https://git.company.somewhere/api/v4/projects/3319/merge_requests/507/discussions: 500 {message: 500 Internal Server Error}

As it is a company managed gitlab, I do not have access to its logs to see what the internal server error is.
The gitlab version is: GitLab Community Edition v17.5.5

My account is a managed ldap account.

Logs from the gitlab.nvim.log file:

`
-- REQUEST TO GITLAB --
POST /api/v4/projects/3319/merge_requests/507/discussions HTTP/1.1
Host: company.somewhere
Accept: application/json
Content-Type: application/json
Private-Token: REDACTED
User-Agent: go-gitlab

{"body":"test","position":{"base_sha":"7a0e2ade58e98f04ac28c7e487d43d18178bcab2","head_sha":"9176a8a2f380184310ddb2484d529c251449d92b","start_sha":"7a0e2ade58e98f04ac28c7e487d43d18178bcab2","new_path":"file/pom.xml","old_path":"","position_type":"text","old_line":66}}

-- RESPONSE FROM GITLAB --
HTTP/1.1 500 Internal Server Error
Content-Length: 39
Cache-Control: no-cache
Connection: keep-alive
Content-Type: application/json
Date: Thu, 20 Mar 2025 10:56:06 GMT
Server: nginx
Strict-Transport-Security: max-age=63072000
Vary: Origin
X-Gitlab-Meta: {"correlation_id":"01JPSK4CXTC5PAC107FVWY4KMG","version":"1"}
X-Request-Id: 01JPSK4CXTC5PAC107FVWY4KMG
X-Runtime: 0.132080

{"message":"500 Internal Server Error"}

-- RESPONSE FROM GO SERVER --
HTTP/0.0 500 Internal Server Error
Content-Length: 173
Content-Type: application/json

[...]

PR #485, opened 2025-03-02
Fix: Swap file_name and old_file_name in reviewer data

jakubbortlik:
This PR fixes #486 by restoring how old paths for (potentially) renamed files are handled. This requires rebuilding the Go server.

This PR fixes also the correct OLD/NEW file name assignment in the reviewer data and also the way how the file name for the comment popup title is created based on the reviewer data.

PR #489, opened 2025-04-03
Fix: Internal server error

jakubbortlik:
This is basically a duplicate of #485, but rebased onto `main`, in case you (@harrisoncramer) want to merge it directly without going through `develop`. I'm not sure which way you prefer, but I'll be mostly unavailable for the rest of the week and I don't want to keep the users waiting for the fix (and I think it would be a pity to just revert the `jump-to-reviewer` fix).

* fix: use new file name if old file name is empty (#486)
* fix: swap `file_name` and `old_file_name` in reviewer data


harrisoncramer:
I don't want to merge something directly into main, I've merged #485 instead.

PR #491, opened 2025-04-10
Feat: Enable toggling date format in discussions

jakubbortlik:
Sometime I find it useful to look at a MR discussion and see when exactly a comment was made. I can of course look at "10 days ago" and then go to the calendar and see what date that was, but that's inconvenient.

This PR makes it possible to toggle between two different formats:
- the current relative format (though this too becomes absolute after 30 days)
- absolute time like "02/28/2025 at 09:30"

This PR suggests a new keybinding `dt` (mnemonic "**d**ate **t**oggle", just like `st` is the default for "**s**ort method **t**oggle").


harrisoncramer:
I just released develop and rebased main onto it, so you may want to merge/rebase develop back into this branch, there are a lot of changes in here that you probably don't intend to make.

Issue #498, opened 2025-06-13
Feature Request: Proxy for GitLab connection


duckbrain:
## Feature Description

I have to use a SOCKS5 proxy to connect to my GitLab instance. 

This can be done by setting `Proxy` on the `net/http.Transport` using either [ProxyFromEnvironment](https://pkg.go.dev/net/http#ProxyFromEnvironment) or [ProxyURL](https://pkg.go.dev/net/http#ProxyURL)

PR #499, opened 2025-06-13
Proxy Support

duckbrain:
Implements #498

### DIFF
--- a/cmd/app/client.go
+++ b/cmd/app/client.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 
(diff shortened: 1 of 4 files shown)


harrisoncramer:
Looks good. Do you mind targeting develop branch instead of main? That's how we do releases (we use develop as a non-stable branch and then release from there to main). Thanks for the contribution!!

duckbrain:
Yep. Looks like main had some commits that weren't on develop. I'll have to rebase when I get to a keyboard.

duckbrain:
Ha, I now see #502 to bring that up-to-date. Rebase is done anyway.

harrisoncramer:
Thank you for the contribution! It'll go out with the next release.

PR #504, opened 2025-08-06
Draft: Feat: Add suggestions functionality

jakubbortlik:
Hi @harrisoncramer, in this rather large PR, I would like to introduce suggestions functionality in the discussion tree and in the reviewer:
- preview a suggestion in a new diff tab (i.e., apply a suggestion and possibly save modifications to local file)
- edit a comment with a suggestion in a new diff tab
- reply to a comment with a suggestion preview
- create a new comment with a suggestion preview

The suggestion preview has some useful features:
- keybinding to paste in a suggestion snippet
- autocommands, diagnostics and signs that make it easy to preview multiple suggestions in one comment

I've tried to make it so that the suggestion is previewed on the local file in as many cases as possible so the user can make use of LSP, running tests with the suggestion applied, etc.

Documentation in `doc/gitlab.nvim.txt` is largely missing yet (that's why the PR is in Draft mode), but the code is fairly well documented.

I'd appreciate feedback, both on the code as well as on the functionality. I've been using this for some time and find it useful and worth the large amount of changes.

I believe some refactoring could be applied to the [existing comment creation](https://github.com/harrisoncramer/gitlab.nvim/blob/b12236f38440bd0e78fcc0b2e17c55f8d911a9dc/lua/gitlab/actions/comment.lua#L209-L296) which would ultimately reduce the amount of code, like I've written a [separate function for creating a suggestion](https://github.com/jakubbortlik/gitlab.nvim/blob/d870050a6b39ad8489ce1599f5ed7212c5774413/lua/gitlab/actions/suggestions.lua#L284-L301), which is a little more universal than the existing [build_suggestion](https://github.com/harrisoncramer/gitlab.nvim/blob/b12236f38440bd0e78fcc0b2e17c55f8d911a9dc/lua/gitlab/actions/comment.lua#L241-L274).


doc/gitlab.nvim.txt
```diff
@@ -223,11 +223,21 @@ you call this function with no values the defaults will be used:
...
+          discard_changes = "ZQ", -- Close suggestion preview tab and discard changes to local file
+          attach_file = "ZA", -- Attach a file from the `settings.attachment_dir`
+          paste_default_suggestion = "glS", -- Paste the default suggestion below the cursor (overrides default "glS" (start review) keybinding for the "Note" buffer)
         },
         reviewer = {
           disable_all = false, -- Disable all default mappings for the reviewer windows
           create_comment = "c", -- Create a comment for the lines that the following {motion} moves over. Repeat the key(s) for creating comment for the current line
           create_suggestion = "s", -- Create a suggestion for the lines that the following {motion} moves over. Repeat the key(s) for creating comment for the current line
+          create_suggestion_with_preview = "S", -- In a new tab create a suggestion with a diff preview for the lines that the following {motion} moves over. Repeat the key(s) for creating comment for the current line
```
harrisoncramer, line 240: This shouldn't be just "S" since that's a native Vim binding and won't trigger right away.

lua/gitlab/actions/discussions/init.lua
```diff
@@ -240,12 +239,84 @@ M.reply = function(tree)
...
     file_name = discussion_node.file_name,
   })
 
   layout:mount()
 end
 
+---Open a new tab with a suggestion preview.
+---@param tree NuiTree The current discussion tree instance.
+---@param action "reply"|"edit"|"apply" Reply to the current thread, edit the current comment or apply the suggestion to local file.
+M.suggestion_preview = function(tree, action)
```
harrisoncramer, line 252: This isn't working for me, the suggestion is hidden, and slowly expands each time I press a character 

https://github.com/user-attachments/assets/796f6bed-d4ad-44c2-a626-00e28472d554

doc/gitlab.nvim.txt
```diff
@@ -223,11 +223,21 @@ you call this function with no values the defaults will be used:
           toggle_unresolved_discussions = "U", -- Open or close all unresolved discussions
           refresh_data = "<C-R>", -- Refresh the data in the view by hitting Gitlab's APIs again
           print_node = "<leader>p", -- Print the current node (for debugging)
+          edit_suggestion = "se", -- Edit comment with suggestion preview in a new tab
+          reply_with_suggestion = "sr", -- Reply to comment with a suggestion preview in a new tab
+          apply_suggestion = "sa", -- Apply the suggestion to the local file with a preview in a new tab
+        },
+        suggestion_preview = {
+          apply_changes = "ZZ", -- Close suggestion preview tab, and post suggestion comment to Gitlab (and discard changes to local file) or "apply" changes to local file
```
harrisoncramer, line 231: Was getting these errors sometimes on submit:

```
gitlab.nvim: Could not create discussion: POST https://gitlab.com/api/v4/projects/45056705/merge_requests/49/discussions: 400 {message: 400 Bad request - Note {:line_code=>["can't be blank", "must be a valid line code"]}}
```

lua/gitlab/actions/discussions/init.lua
```diff
@@ -240,12 +239,84 @@ M.reply = function(tree)
...
+    is_new_sha = is_new_sha,
+    revision = revision,
+    note_header = note_node.text,
+    comment_type = is_draft and "draft" or action,
+    note_lines = action ~= "reply" and common.get_note_lines(tree) or nil,
+    root_node_id = root_node.id,
+    note_node_id = note_node_id,
+  }
+  require("gitlab.actions.suggestions").show_preview(opts)
+end
```
harrisoncramer, line 318: This is confusing to me. Does "apply" not actually apply the suggestion? All it's doing for me locally is opening the preview. I'd expect the behavior would be to apply the suggestion to the MR, as it does in Gitlab, and like close the issue as well. Is that not possible?

lua/gitlab/actions/discussions/init.lua
jakubbortlik: I remember that I had the behaviour you describe before but then I added a [fix](https://github.com/jakubbortlik/gitlab.nvim/blob/8ea62ef822714125fb7b061fab45fbcfb6b6d908/lua/gitlab/actions/suggestions.lua#L42-L46) for that, so maybe you need to pull the latest `feat-add-suggestions-functionality` branch.

I'm still getting a similar behaviour when I leave the suggestion intact and only modify the comment above the" \`\`\`suggestion:-0+0" line. But as soon as I make a modification in the suggestion body itself, the diff is shown correctly. It would be best to fix this as well, I've experimented with different folding normal mode commands instead of [zX](https://github.com/jakubbortlik/gitlab.nvim/blob/8ea62ef822714125fb7b061fab45fbcfb6b6d908/lua/gitlab/actions/suggestions.lua#L45), but haven't found a solution yet.

lua/gitlab/actions/discussions/init.lua
```diff
@@ -240,12 +239,84 @@ M.reply = function(tree)
...
+    is_new_sha = true
+  end
+
+  -- Get values for preview depending on whether comment is on OLD or NEW version
+  local revision
+  if is_new_sha then
+    revision = common.commented_line_has_changed(tree, root_node) and root_node.head_sha or "HEAD"
+  else
+    revision = root_node.base_sha
+  end
```
jakubbortlik, line 301: If the comment is on the "NEW SHA", we can use HEAD as the original text of the suggestion and show it on the local file. Othewise we use the `base_sha` (i.e. the version in the target branch) and show it in a temp file. This reduces the usefulness of the suggestion preview as it doesn't allow for using all of the LSP goodies etc.

For this reason, I'd like to reverse the logic of [showing comments on unmodified lines in the old file](https://github.com/jakubbortlik/gitlab.nvim/blob/c7ba8b776dbf4e1f230c3dd4e39e8be22e72fc45/lua/gitlab/reviewer/init.lua#L194-L196) and show them in the new file whenever possible. What do you think?

lua/gitlab/actions/discussions/init.lua
jakubbortlik: I've just pushed a naive implementation that applies & commits the suggestion, pushes to remote and resolves the thread.

It would be nice if the user could modify the commit message but I didn't find a simple way of doing that.

I'd also like to make it possible to create suggestion batches, but that would be even more complicated.

lua/gitlab/actions/suggestions.lua
```diff
@@ -0,0 +1,732 @@
...
+        elseif opts.comment_type == "edit" then
+          require("gitlab.actions.comment").confirm_edit_comment(opts.root_node_id, opts.note_node_id, false)(buf_text)
+        elseif opts.comment_type == "new" then
+          require("gitlab.actions.comment").confirm_create_comment(buf_text, false)
+        elseif opts.comment_type == "apply" then
+          if imply_local then
+            original_lines = vim.api.nvim_buf_get_lines(suggestion_buf, 0, -1, false)
+            if
+              not git.add({ filename = vim.fn.bufname(suggestion_buf) })
+              or not git.commit({ commit_message = "Apply 1 suggestion to 1 file" })
```
jakubbortlik, line 147: This naive approach is really problematic because it doesn't consider the possibility that the user has already staged some changes and the commit would contain those changes.

I'll explore the possibility of using the [CreateCommit API](https://gitlab.com/gitlab-community/gitlab-org/api/client-go/-/blob/e1b29adfd37db39aae4e1547f336b71d67efcdb8/commits.go#L248) for this, instead.

PR #528, opened 2026-02-12
feat: better handling of server installation path

acristoffers:
Closes #493

I also use Nix to configure my NeoVim installation. The modifications in this MR make it possible to use the plugin this way, or with another form of distribution that does not provide git or has a read-only plugin folder.

This is achieved by:

1. Instead of hard-coding the server binary path to the plugin folder, use `vim.fn.stdpath("data")`. This is the correct place to put it in. But anyway, don't assume it's writable, and try it and all folders in the runtime to find one that we can write to, and put it there. The data folder is most probably going to be writable, but this way we can at least have more options before failing. This is already done by NeoVim itself when you try to install a spell dictionary, for example, or a tree-sitter grammar, so it is a safe way to handle this that aligns with existing practices.
2. Give the user the option of pointing to an existing server, because why not?

To this end I changed the configuration structure to have a `server` key with a `binary` sub-key, and moved the `port` option inside it, for coherence.

The init script was changed to set the configuration before building, as we now need the configuration during the build.

The build function was, of course, the most affected by the changes.

A `flake.nix` is provided that make it easy to use the plugin with nix. It offers the plugin and server, so the user can decide if they want to let the plugin build the server as usual or if they want to let nix manage the build and just point to it in their configuration.

The server assumes the path of the emoji configuration file. I don't understand why the file is provided like that, could it not be embedded in the binary? Maybe its path could be passed as configuration by Lua when launching it? I decided to not change it though, and just make sure the config folder is where the server expects it to be.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,47 @@
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.11";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, flake-utils, nixpkgs }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; config.allowUnfree = true; };
+        gitlab-nvim-server = pkgs.buildGoModule {
+          pname = "gitlab.nvim-server";
+          version = "git";
+          src = ./.;
+          vendorHash = "sha256-OLAKTdzqynBDHqWV5RzIpfc3xZDm6uYyLD4rxbh0DMg=";
+          postInstall = ''
+            cp -r ${./cmd/config} $out/bin/config
+            mv $out/bin/cmd $out/bin/gitlab.nvim
+          '';
+        };
+        gitlab-nvim = pkgs.vimUtils.buildVimPlugin {
+          name = "gitlab.nvim";
+          src = ./.;
+          doCheck = false;
+        };
+      in
+      rec {
+        formatter = pkgs.nixpkgs-fmt;
+        packages.gitlab-nvim-server = gitlab-nvim-server;
+        packages.gitlab-nvim = gitlab-nvim;
+        packages.default = packages.gitlab-nvim;
+        devShell = pkgs.mkShell {
+          packages = with pkgs; [
+            git
+            go
+            go-tools
+            golangci-lint
+            luajitPackages.busted
+            luajitPackages.luacheck
+            luajitPackages.luarocks
+            neovim
+            stylua
+          ];
+        };
+      }
+    );
+}
--- a/flake.nix
+++ b/flake.nix
@@ -7,7 +7,7 @@
   outputs = { self, flake-utils, nixpkgs }:
     flake-utils.lib.eachDefaultSystem (system:
       let
-        pkgs = import nixpkgs { inherit system; config.allowUnfree = true; };
+        pkgs = import nixpkgs { inherit system; };
         gitlab-nvim-server = pkgs.buildGoModule {
           pname = "gitlab.nvim-server";
           version = "git";
@@ -24,12 +24,12 @@
           doCheck = false;
         };
       in
-      rec {
+      {
         formatter = pkgs.nixpkgs-fmt;
         packages.gitlab-nvim-server = gitlab-nvim-server;
         packages.gitlab-nvim = gitlab-nvim;
-        packages.default = packages.gitlab-nvim;
-        devShell = pkgs.mkShell {
+        packages.default = gitlab-nvim;
+        devShells.default = pkgs.mkShell {
           packages = with pkgs; [
             git
             go
--- a/flake.nix
+++ b/flake.nix
@@ -2,17 +2,25 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.11";
     flake-utils.url = "github:numtide/flake-utils";
+    gomod2nix = {
+      url = "github:nix-community/gomod2nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
   };
 
-  outputs = { self, flake-utils, nixpkgs }:
+  outputs = { self, flake-utils, nixpkgs, gomod2nix }:
     flake-utils.lib.eachDefaultSystem (system:
       let
-        pkgs = import nixpkgs { inherit system; };
-        gitlab-nvim-server = pkgs.buildGoModule {
+        pkgs = import nixpkgs {
+          inherit system;
+          overlays = [ gomod2nix.overlays.default ];
+        };
+        gitlab-nvim-server = pkgs.buildGoApplication {
           pname = "gitlab.nvim-server";
           version = "git";
           src = ./.;
-          vendorHash = "sha256-OLAKTdzqynBDHqWV5RzIpfc3xZDm6uYyLD4rxbh0DMg=";
+          modules = ./gomod2nix.toml;
+          subPackages = [ "cmd" ];
           postInstall = ''
             cp -r ${./cmd/config} $out/bin/config
             mv $out/bin/cmd $out/bin/gitlab.nvim
@@ -34,6 +42,7 @@
             git
             go
             go-tools
+            gomod2nix
             golangci-lint
             luajitPackages.busted
             luajitPackages.luacheck
--- a/lua/gitlab/job.lua
+++ b/lua/gitlab/job.lua
@@ -7,10 +7,6 @@ local M = {}
 M.run_job = function(endpoint, method, body, callback, on_error_callback)
   local state = require("gitlab.state")
   local port = state.settings.server and state.settings.server.port
-  if state.settings.port ~= nil then
-    port = state.settings.port
-    u.notify("The setting `port` has been renamed `server.port`", vim.log.levels.WARN)
-  end
   local args = { "-s", "-X", (method or "POST"), string.format("localhost:%s%s", port, endpoint) }
 
   if body ~= nil then
--- a/lua/gitlab/job.lua
+++ b/lua/gitlab/job.lua
@@ -6,7 +6,12 @@ local M = {}
 
 M.run_job = function(endpoint, method, body, callback, on_error_callback)
   local state = require("gitlab.state")
-  local args = { "-s", "-X", (method or "POST"), string.format("localhost:%s", state.settings.port) .. endpoint }
+  local port = state.settings.server and state.settings.server.port
+  if state.settings.port ~= nil then
+    port = state.settings.port
+    u.notify("The setting `port` has been renamed `server.port`", vim.log.levels.WARN)
+  end
+  local args = { "-s", "-X", (method or "POST"), string.format("localhost:%s%s", port, endpoint) }
 
   if body ~= nil then
     local encoded_body = vim.json.encode(body)
--- a/lua/gitlab/server.lua
+++ b/lua/gitlab/server.lua
@@ -30,7 +30,7 @@ end
 
 -- Starts the Go server and call the callback provided
 M.start = function(callback)
-  local port = tonumber(state.settings.port) or 0
+  local port = tonumber(state.settings.server.port) or 0
   local parsed_port = nil
   local callback_called = false
 
@@ -51,7 +51,7 @@ M.start = function(callback)
     settings = settings:gsub('"', '\\"')
   end
 
-  local command = string.format('"%s" "%s"', state.settings.bin, settings)
+  local command = string.format('"%s" "%s"', state.settings.server.binary, settings)
 
   local job_id = vim.fn.jobstart(command, {
     on_stdout = function(_, data)
@@ -61,7 +61,7 @@ M.start = function(callback)
           port = line:match("Server started on port:%s+(%d+)")
           if port ~= nil then
             parsed_port = port
-            state.settings.port = port
+            state.settings.server.port = port
             break
           end
         end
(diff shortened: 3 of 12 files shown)


lua/gitlab/state.lua
```diff
@@ -45,7 +45,10 @@ end
 M.settings = {
   auth_provider = M.default_auth_provider,
   file_separator = u.path_separator,
-  port = nil, -- choose random port
+  server = {
+    binary = nil, -- path to the server binary. if nil, it will be compiled, otherwise the existing binary at this path will be used
+    port = nil, -- choose random port
```
jakubbortlik: I think it's best to leave any documentation to just the `doc/gitlab.nvim.txt` file.

acristoffers: Comments removed.

acristoffers:
I forgot to add a shell to the nix flake, so I added it and force pushed. It allows developers to have a shell with all dependencies installed with one command.

To test the environment I tried running the tests and run into a problem: the test expects a date string back in English, but my locale is set to Portuguese, so the strings didn't match. I modified the `lua-test.sh` script to set `LC_TIME` to English before testing, so it doesn't fail because of the user environment.

lua/gitlab/job.lua
```diff
@@ -6,7 +6,7 @@ local M = {}
 
 M.run_job = function(endpoint, method, body, callback)
   local state = require("gitlab.state")
-  local args = { "-s", "-X", (method or "POST"), string.format("localhost:%s", state.settings.port) .. endpoint }
+  local args = { "-s", "-X", (method or "POST"), string.format("localhost:%s", state.settings.server.port) .. endpoint }
```
harrisoncramer: The only thing here I'd recommend is that for backward compatibility reasons we also just set this as state.settings.port if it's around. That way people who previously set this value don't have their configurations broken.

harrisoncramer (dismissed):

lua/gitlab/job.lua
jakubbortlik: Maybe we could add a warning that the setting has changed and `state.settings.port` will be removed in a future version. The healthcheck already checks [removed settings](https://github.com/harrisoncramer/gitlab.nvim/blob/3d2828a9504b87fc36ee2aca1b0f36cf75003edd/lua/gitlab/health.lua#L70-L97), so maybe it could also report *deprecated* settings. This way we could in the future get rid of special handling of deprecated settings like the [merge_status](https://github.com/harrisoncramer/gitlab.nvim/blob/6ed09564b70a65750281c5540729a30f87f4ea13/lua/gitlab/actions/summary.lua#L172-L174) key in `state.settings.info.fields`.

acristoffers: I amended with the changes. It now takes `settings.port` if it is set, but emits a warning.

lua/gitlab/job.lua
```diff
@@ -4,9 +4,14 @@ local Job = require("plenary.job")
...
 
-M.run_job = function(endpoint, method, body, callback, on_error_callback)
+M.run_job = function(endpoint, method, body, callback)
   local state = require("gitlab.state")
-  local args = { "-s", "-X", (method or "POST"), string.format("localhost:%s", state.settings.port) .. endpoint }
+  local port = state.settings.server and state.settings.server.port
+  if state.settings.port ~= nil then
+    port = state.settings.port
+    u.notify("The setting `port` has been renamed `server.port`", vim.log.levels.WARN)
+  end
```
jakubbortlik: Hi @acristoffers this is not the right place to fall back to `sate.settings.port`. This has to be done in `lua/gitlab/server.lua`, otherwise if the user doesn't set `settings.server.port` it is assigned automatically and then it clashes with `state.settings.port` and I get `gitlab.nvim: Go server exited with non-zero code: 7`.

lua/gitlab/server.lua
```diff
@@ -105,26 +105,59 @@ end
 -- Builds the Go binary with the current Git tag.
 M.build = function(override)
   local file_path = u.current_file_path()
-  local parent_dir = vim.fn.fnamemodify(file_path, ":h:h:h:h")
+  state.settings.root_path = vim.fn.fnamemodify(file_path, ":h:h:h:h")
```
jakubbortlik: You could just as well fix the `u.current_file_path()` function to actually return the root path directly. It is only used in the `server.lua` file but it doesn't really return the "current file path" but the path of the `utils.lua`. So if you modified the function to return the root, you could just use it like this:
```diff
diff --git a/lua/gitlab/server.lua b/lua/gitlab/server.lua
index 55f04de..0cfbe5c 100644
--- a/lua/gitlab/server.lua
+++ b/lua/gitlab/server.lua
@@ -104,8 +104,7 @@ end
 
 -- Builds the Go binary with the current Git tag.
 M.build = function(override)
-  local file_path = u.current_file_path()
-  state.settings.root_path = vim.fn.fnamemodify(file_path, ":h:h:h:h")
+  state.settings.root_path = u.get_root_path()
 
   -- If the user provided a path to the server, don't build it.
   if state.settings.server.binary ~= nil then
```

jakubbortlik: Then of course, you'd have to also change the code in the [get_version](https://github.com/harrisoncramer/gitlab.nvim/blob/eb73c8cde9fc589388b2f91ae8f7ea3d97bdaeb7/lua/gitlab/server.lua#L221-L222) function.

flake.nix
```diff
@@ -0,0 +1,47 @@
...
+          name = "gitlab.nvim";
+          src = ./.;
+          doCheck = false;
+        };
+      in
+      rec {
+        formatter = pkgs.nixpkgs-fmt;
+        packages.gitlab-nvim-server = gitlab-nvim-server;
+        packages.gitlab-nvim = gitlab-nvim;
+        packages.default = packages.gitlab-nvim;
```
jakubbortlik: I don't use `nix` myself, but Claude says to this that the `rec` keyword is not necessary if you define the `packages.default` as follows (without the recursion):
```suggestion
      {
        formatter = pkgs.nixpkgs-fmt;
        packages.gitlab-nvim-server = gitlab-nvim-server;
        packages.gitlab-nvim = gitlab-nvim;
        packages.default = gitlab-nvim;
```

flake.nix
```diff
@@ -0,0 +1,47 @@
...
+        devShell = pkgs.mkShell {
```
jakubbortlik: Claude also says, the `devShell` is deprecated in favour of `devShells.default`:
```suggestion
        devShells.default = pkgs.mkShell {
```

flake.nix
```diff
@@ -0,0 +1,47 @@
...
+
+  outputs = { self, flake-utils, nixpkgs }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; config.allowUnfree = true; };
+        gitlab-nvim-server = pkgs.buildGoModule {
+          pname = "gitlab.nvim-server";
+          version = "git";
+          src = ./.;
+          vendorHash = "sha256-OLAKTdzqynBDHqWV5RzIpfc3xZDm6uYyLD4rxbh0DMg=";
```
jakubbortlik: Claude came up with this:
```
2. Hardcoded vendorHash (line 15)
- The hash sha256-OLAKTdzqynBDHqWV5RzIpfc3xZDm6uYyLD4rxbh0DMg= will break when Go
dependencies change
```
Does that mean that whenever the Go dependencies change this `flake.nix` file will have to be updated to keep working? How is the `vendorHash` created? Who will be responsible for updating it?

flake.nix
```diff
@@ -0,0 +1,47 @@
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.11";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, flake-utils, nixpkgs }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; config.allowUnfree = true; };
```
jakubbortlik: Is the `config.allowUnfree = true` necessary? What Claude says to this:
```
  Which Packages Here Are Unfree?

  Looking at the flake, I don't actually see any obviously unfree packages:

  # devShell packages (line 33-42)
  git, go, go-tools, golangci-lint, luajitPackages.*, neovim, stylua

  These are all open source. The allowUnfree = true may be:
  1. Legacy/copy-pasted from another flake
  2. Added "just in case" during development
  3. No longer needed

  You could likely remove it entirely and see if the build still works.

```
So, is the setting really necessary?

lua/gitlab/server.lua
```diff
@@ -133,6 +166,12 @@ M.build = function(override)
     u.notify(string.format("Failed to install with status code %d:\n%s", res.code, res.stderr), vim.log.levels.ERROR)
     return false
   end
+
+  local Path = require("plenary.path")
+  local src = Path:new(state.settings.root_path .. u.path_separator .. "cmd" .. u.path_separator .. "config")
+  local dest = Path:new(bin_folder .. u.path_separator .. "config")
+  src:copy({ destination = dest, recursive = true, override = true })
```
jakubbortlik, line 176: This is copied because of the code in [cmd/app/emoji.go](https://github.com/harrisoncramer/gitlab.nvim/blob/3d2828a9504b87fc36ee2aca1b0f36cf75003edd/cmd/app/emoji.go#L149-L182), right? As far as I can see, this code does nothing. The emoji map is only needed for the lua code, so unless @harrisoncramer decides otherwise, the `attachEmojis` function and it's [usage](https://github.com/harrisoncramer/gitlab.nvim/blob/3d2828a9504b87fc36ee2aca1b0f36cf75003edd/cmd/app/server.go#L25-L32) can be deleted. Then we won't need to copy the config folder.

Even if you don't delete `attachEmojis` function, an alternative to copying the config would be to pass the path to `emojis.json` as part of the [PluginOptions](https://github.com/harrisoncramer/gitlab.nvim/blob/3d2828a9504b87fc36ee2aca1b0f36cf75003edd/cmd/app/config.go#L3) when the [server is started](https://github.com/harrisoncramer/gitlab.nvim/blob/eb73c8cde9fc589388b2f91ae8f7ea3d97bdaeb7/lua/gitlab/server.lua#L37-L45).

lua/gitlab/job.lua
jakubbortlik: This still needs addressing, but I've created a [new comment](https://github.com/harrisoncramer/gitlab.nvim/pull/528/changes/BASE..829cefd0ac014af22d4afc9f3f9afad7d66b59f1#r2999490165) for the amended code, so closing this thread.

acristoffers:
I pushed changes as a `fixup!` commit, so don't merge yet, unless you have squashing enabled.

lua/gitlab/job.lua
acristoffers: Change made.

lua/gitlab/server.lua
acristoffers: Changed.

flake.nix
acristoffers: Doesn't really matter, but I modified it as requested.

flake.nix
acristoffers: I not even knew this was changed... I modified it (and will check all my projects :upside_down_face: )

flake.nix
acristoffers: Yes, unfortunately. But only when there are dependencies changes, not code changes, so not so often. To generate it, I edit the nix file and remove the hash (it becomes an empty string), then I run `nix build` and it will fail and say what hash is, which I then copy into the file.

I can keep it up-to-date, as I'll quickly see the problem (this is in my NeoVim flake, and I rebuild my flake every day).

flake.nix
acristoffers: I guess it is a leftover from a copy. I don't write flakes from scratch :P I copy from other projects (mine) and modify, but I overlooked it. I fixed it.

lua/gitlab/server.lua
acristoffers: I don't want to make design decisions, so I'll wait for @harrisoncramer to decide.

flake.nix
acristoffers: I personally prefer to update the hashes manually, but there is a solution that allows a GitHub action to do it. Since I guess that you as project would prefer it to be automated, I pushed another `fixup!` commit that makes use of that. It introduces a TOML file that holds dependencies hashes and a GitHub action that updates this file and opens a PR with the changes.

lua/gitlab/utils/init.lua
```diff
@@ -509,6 +509,12 @@ M.current_file_path = function()
   return vim.fn.fnamemodify(path, ":p")
 end
```
jakubbortlik: @acristoffers I think you can safely remove this function. It's not used anywhere now and I don't think it is particularly useful since it's tied to the utils init file.

flake.nix
jakubbortlik: I leave this up to Harrison, if he wants to have this in the plugin.

jakubbortlik:
@acristoffers could you please rebase? The develop branch has been updated.

lua/gitlab/utils/init.lua
acristoffers: Done.

jakubbortlik:
Hi @acristoffers. I've just realized, you also should update the new structure of the settings in [annotations.lua](https://github.com/acristoffers/gitlab.nvim/blob/9f898aa1a8cd74fc11756c295a56ea0d4952cf40/lua/gitlab/annotations.lua#L149).

acristoffers:
> Hi @acristoffers. I've just realized, you also should update the new structure of the settings in [annotations.lua](https://github.com/acristoffers/gitlab.nvim/blob/9f898aa1a8cd74fc11756c295a56ea0d4952cf40/lua/gitlab/annotations.lua#L149).

Done

PR #530, opened 2026-02-18
feat!: implement AutoMerge option to AcceptMergeRequest

jakubbortlik:
Closes #518.

This PR also fixes a bug when the `Delete source branch` setting of the MR was not respected due to incorrect reference in `lua/gitlab/actions/merge.lua` to `state.INFO.delete_branch` instead of `state.INFO.force_remove_source_branch`.

This PR also changes the behaviour of how the `opts` in `gitlab.merge(opts)` behave:

- Before this PR:
  - when `lua require("gitlab").merge()` was run without parameters, the values visible in the Summary view were used (except for the bug mentioned above).
  - specifying one of `squash` or `delete_branch` in `opts` caused the other parameter to default to `false` even if Summary view would show it's set to `true`.

- After this PR:
  - running `lua require("gitlab").merge()` without parameters doesn't change.
  - specifying any of the `opts` values (`auto_merge`, `squash`, `delete_branch`), has no effect on the other options - if not specified in the `gitlab.merge()` call, the values from the Summary are used.

This is a breaking change since before, a user could rely on the fact that ignoring one of the options would set it to `false` which now is not guaranteed - the value depends on the existing MR settings, which I find to be the desirable behaviour:
- only override what needs to be overridden
- only one set of default values (those in the Summary)

### DIFF
--- a/lua/gitlab/actions/merge.lua
+++ b/lua/gitlab/actions/merge.lua
@@ -12,19 +12,23 @@ local function create_squash_message_popup()
 end
 
 ---@class MergeOpts
+---@field auto_merge boolean?
 ---@field delete_branch boolean?
 ---@field squash boolean?
 ---@field squash_message string?
 
 ---@param opts MergeOpts
 M.merge = function(opts)
-  local merge_body = { squash = state.INFO.squash, delete_branch = state.INFO.delete_branch }
-  if opts then
-    merge_body.squash = opts.squash ~= nil and opts.squash
-    merge_body.delete_branch = opts.delete_branch ~= nil and opts.delete_branch
+  local merge_body = {
+    auto_merge = state.INFO.merge_when_pipeline_succeeds,
+    squash = state.INFO.squash,
+    delete_branch = state.INFO.force_remove_source_branch,
+  }
+  for key, val in pairs(opts or {}) do
+    merge_body[key] = val
   end
 
-  if state.INFO.detailed_merge_status ~= "mergeable" then
+  if state.INFO.detailed_merge_status ~= "mergeable" and not merge_body.auto_merge then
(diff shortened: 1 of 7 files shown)


lua/gitlab/actions/merge.lua
jakubbortlik, line 45: I'm thinking about adding this `vim.ui.select` call with a recursive call to `M.merge`, which would make it simpler for users to set `auto-merge`, but maybe it's a pattern that you (@harrisoncramer) don't want to introduce in the codebase.
```lua
  if state.INFO.detailed_merge_status ~= "mergeable" and not merge_body.auto_merge then
    vim.ui.select({ "Set auto-merge", "Cancel" }, {
      prompt = string.format(
        "MR not mergeable (currently '%s'). Set auto-merge instead?",
        state.INFO.detailed_merge_status
      ),
    }, function(choice)
      if choice == "Set auto-merge" then
        local opts_ = opts or {}
        opts_.auto_merge = true
        M.merge(opts_)  -- Recurse here with overridden `auto_merge`
      end
    end)
  else
    if merge_body.squash then
      local squash_message_popup = create_squash_message_popup()
      popup.set_up_autocommands(squash_message_popup, nil, vim.api.nvim_get_current_win())
      squash_message_popup:mount()
      popup.set_popup_keymaps(squash_message_popup, function(text)
        M.confirm_merge(merge_body, text)
      end, nil, popup.editable_popup_opts)
    else
      M.confirm_merge(merge_body)
    end
  end
```

harrisoncramer:
Hey @jakubbortlik this looks good, will merge this in once you resolve conflicts, sorry for the delay here

jakubbortlik:
Hi Harrison, I appreciate you still maintain the plugin even if, as I understand, you don't use it actively anymore.

PR #531, opened 2026-02-24
fix: use plain substring matching with labels

jakubbortlik:
Using `plain` substring matching makes it possible to highlight labels that contain hyphens (or any other "magic" characters).


jakubbortlik:
Superceded by #532.

Issue #444, opened 2024-12-10
Show more details about approvals in MR Summary


jakubbortlik:
## Feature Description
One thing that is missing in `gitlab.nvim` and for which I frequently visit Gitlab online is a more detailed listing of approvals of a MR. In some projects we have rules like two approvals for certain files/filetypes and `gitlab.nvim` only shows `not_approved` in case there is already one approval.

I've just realized, alternatively I could use the `glab` tool for listing this from inside Neovim:
```
:!glab mr approvers

Listing merge request !454 eligible approvers:
Approval rules overwritten.
Rule "All Members" sufficient approvals (1/1 required):
Jakub Bortlík   jakub.bortlik   👍

Rule "*.py" insufficient approvals (1/2 required):
Some Guy   some.guy   -
Jakub Bortlík   jakub.bortlik   👍
```
It would be nice if this information was in some form available in the MR summary in `gitlab.nvim`.

harrisoncramer:
Hey Jakub, this is an interesting idea, thanks for sharing -- you have a few other open issues and feature requests. Do you mind prioritizing those first?

jakubbortlik:
Hi Harrison, yeah this would be nice to have but since there is a workaround with `glab mr approvers` it's low priority. I'd really appreciate if #425 could be fixed - currently I just don't bulk-publish drafts because I don't want to deal with the possible mess it can create.

jakubbortlik:
For anybody who'd like to take a look at this: Getting the approvals from Gitlab is possible with [GetConfiguration](https://gitlab.com/gitlab-community/gitlab-org/api/client-go/-/blob/856a6fcf441cf8783868070291a4e55ff97d3ee1/merge_request_approvals.go#L221-244).

PR #532, opened 2026-02-26
feat: add mergeability checks to summary view

### LINKED ISSUE #508: More detailed meargability tests
## Problem: 
The [detailed_merge_status](https://github.com/harrisoncramer/gitlab.nvim/blob/9f898aa1a8cd74fc11756c295a56ea0d4952cf40/lua/gitlab/actions/summary.lua#L119-L119) in the MR status can only show a single value. This creates a little inconvenience when the MR has both the `not_approved` and `need_rebase` attributes. The `not_approved` takes precedence and the user may take action to ask for approvals. When necessary approvals are granted, the `need_rebase` value is shown and after rebasing, the approvals may be removed requiring another round of approvals that could have been avoided if the `need_rebase` populated the value of `detailed_merge_status` with higher priority than `not_approved`.

## Possible solution
Unfortunately, there's no way of doing this with the REST API, but it would be possible with the [GraphQL API](https://gitlab.com/gitlab-org/gitlab/-/issues/570458#note_2759863014) which is available as an experimental API in the new [client-go](https://gitlab.com/gitlab-community/gitlab-org/api/client-go/-/blob/b19cd7c43b7b129f9eef94badf3687084348f2e6/graphql.go#L62-L101) library. We could add a GraphQL call to get the results of all the mergegility chechs and show them in the MR status.

Alternatively, we could get all the [MR info data](https://github.com/jakubbortlik/gitlab.nvim/blob/9f898aa1a8cd74fc11756c295a56ea0d4952cf40/lua/gitlab/actions/summary.lua#L114-L114) via GraphQL. I've never worked with GraphQL however, so I'm not sure what that would actually entail.

jakubbortlik:
Closes #508.

This PR uses Gitlab's GraphQL API to retrieve the mergeability checks from the server and shows them in the Summary view:
image

Users can configure it to:
- filter out checks by their status (e.g., those that are "INACTIVE")
- filter out individual checks
- specify the symbols for individual statuses
- specify the descriptions for individual checks


lua/gitlab/init.lua
```diff
@@ -62,6 +63,7 @@ return {
   setup = setup,
   summary = async.sequence({
     u.merge(info, { refresh = true }),
+    u.merge(mergeability, { refresh = true }),
```
jakubbortlik, line 66: I guess it could be more efficient to fetch all the data in one GraphQL request, or only request the mergeability checks if the user includes them in the `settings.info.fields`.

Ultimately, I'd also like to add to the summary the approvals (#444), which would also be possible with the GraphQL API. Then it would make even more sense to do only one request for the data, or perhaps dynamically create the GraphQL query by only requesting those fields that the user configures.

lua/gitlab/actions/summary.lua
jakubbortlik: By introducing the multiline mergeability checks, this iteration over the fields and line indexing would stop working if the mergeability checks were not the last item in the fields table.

I guess, I should fix this in this PR.

jakubbortlik: I've fixed the highlighting for multiline fields and added some more highlights.

harrisoncramer (approved):

PR #533, opened 2026-03-02
feat: check local branch state when creating comments

jakubbortlik:
Comment creation sometimes crashes when the local branch is not up-to-date with the remote, sometimes a comment is created on the wrong line when the file has been updated on remote.

The out-of-sync state is reported when the reviewer is initiated, but it should be checked more regularly, so this PR introduces checking of "ahead" and "behind" commits:
- whenever a user tries to create a new comment
- when the discussion tree is updated

The "ahead" and "behind" commits are now shown in the winbar of the discussion tree. To make it fit more easily, I've changed the section "Inline Comments" to just "Comments".

### DIFF
--- a/lua/gitlab/git.lua
+++ b/lua/gitlab/git.lua
@@ -74,14 +74,17 @@ M.fetch_remote_branch = function(remote_branch)
   return true
 end
 
----Determines whether the tracking branch is ahead of or behind the current branch, and warns the user if so
----@param current_branch string
----@param remote_branch string
----@param log_level number
----@return boolean
-M.get_ahead_behind = function(current_branch, remote_branch, log_level)
+---Determines whether the tracking branch is ahead of or behind the current branch and returns the
+---number of ahead and behind commits or nil values in case of errors.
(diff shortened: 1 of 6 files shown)


lua/gitlab/git.lua
```diff
@@ -74,54 +74,34 @@ M.fetch_remote_branch = function(remote_branch)
...
----@param remote_branch string
----@param log_level number
----@return boolean
-M.get_ahead_behind = function(current_branch, remote_branch, log_level)
+---Determines whether the tracking branch is ahead of or behind the current branch and returns the
+---number of ahead and behind commits or nil values in case of errors.
+---@param current_branch string|nil
+---@param remote_branch string|nil
+---@return integer|nil ahead, integer|nil behind
+M.get_ahead_behind = function(current_branch, remote_branch)
```
jakubbortlik, line 82: This function is refactored to only get the number of commits and not show warnings to the user so that it can be used more often without annoying the user. The notification is moved up to the `check_current_branch_up_to_date_on_remote` function.

Issue #386, opened 2024-10-04
Trying to make a comment raises the error: line_code=>["can't be blank", "must be a valid line code"]


eonpatapon:
## Prerequsities

- [x] I'm on the latest version of the plugin
- [x] I've installed the required dependencies
- [x] I've run `:h gitlab.nvim.troubleshooting` and followed the steps there

## Setup Configuration and Environment

```
require("gitlab").setup({debug = { go_request = true, go_response = true, request = true, response = true, gitlab_request = true, gitlab_response = true, }})
```

Otherwise, only creds are setup

## Bug Description

I'm using a private gitlab enterprise server and this plugin at `38bde8a0e4a982329ab4dc4f214f3f06dfc4e4fd`

So I'm able to open a review, then in the review diff in visual mode I select one line (same with multiple lines) and try to add a comment with `<leader>c`.

The request is like so:

```
-- REQUEST TO GO SERVER --
POST /mr/comment HTTP/1.1
Host: localhost:38973
Accept: */*
Content-Length: 444
Content-Type: application/x-www-form-urlencoded
Private-Token: REDACTED
User-Agent: curl/8.7.1

{"file_name":"xxxx","base_commit_sha":"ba44dad8a1016eb24b3b28948dbd8be10e51d727","start_commit_sha":"ba44dad8a1016eb24b3b28948dbd8be10e51d727","head_commit_sha":"419920e26de375a8094e269ae55907410a98bc1e","comment":"Test","new_line":309,"old_line":310,"type":"text","line_range":{"end":{"new_line":309,"old_line":310,"type":"old"},"start":{"new_line":309,"old_line":310,"type":"old"}}}

-- REQUEST TO GITLAB --
POST /api/v4/projects/373/merge_requests/747/discussions HTTP/1.1
Host: XXXX
Accept: application/json
Content-Type: application/json
Private-Token: REDACTED
User-Agent: go-gitlab

[...]

PR #536, opened 2026-03-25
Prevent usage of external diff tools

SlayerOfTheBad:
Makes `parse_hunks_and_diff` pass `--no-ext-diff` to `git diff` to prevent an external diff tool configured in `~/.gitconfig` clobering the diff output it depends on to parse the hunks.

This resolves some, but likely not all, instances of https://github.com/harrisoncramer/gitlab.nvim/issues/386


jakubbortlik:
Thanks for your contribution.

PR #546, opened 2026-04-22
feat: Add support for merge request from fork

Clement-Hue:
## Summary

Add support for choosing and reviewing merge requests originating from forked projects. Previously, the plugin assumed the MR source branch always existed in the local repository, which broke fork-based workflows. This MR makes fork MRs discoverable, fetchable, and reviewable via `choose_merge_request`.

## Problem

When a user selects a merge request that originates from a fork, the plugin fails because:
1. The MR middleware always filters by `SourceBranch`, which doesn't exist locally for fork MRs.
2. The pipeline handler cannot resolve the latest commit because the fork branch isn't on the remote.
3. Diffview can't display diffs because the fork commits aren't fetched.
4. The Lua `choose_merge_request` action tries to switch to a branch that doesn't exist locally.

## Changes

- **`cmd/app/git/git.go`** — Add `FetchMRHead` (fetches `refs/merge-requests/<iid>/head`) and `GetMRHeadCommit` (resolves the fetched ref to a SHA) to the `GitManager` interface and `Git` implementation; add `FetchMrHead` helper that skips the fetch when no MR IID is set.
- **`cmd/app/git/git_test.go`** — Add unit tests for `FetchMrHead` (skip/fetch/error paths) and integration tests for `GetMRHeadCommit` (success and non-existent ref) using a temporary git repo.
- **`cmd/app/middleware.go`** — When `ChosenMrIID` is set, filter MRs by IID only (skip `SourceBranch`) so fork MRs are found.
- **`cmd/app/middleware_test.go`** — Add tests verifying the middleware uses `IIDs` vs `SourceBranch` depending on `ChosenMrIID`.
- **`cmd/app/pipeline.go`** — Fall back to `GetMRHeadCommit` when the branch doesn't exist on the remote and an MR IID is set.
- **`cmd/app/test_helpers.go`** — Implement `GetMRHeadCommit` and `FetchMRHead` on the `FakeGitManager` to satisfy the updated interface.
- **`cmd/main.go`** — Call `FetchMrHead` at startup so fork MR commits are available locally for diffview.
- **`lua/gitlab/actions/merge_requests.lua`** — Detect fork MRs (`source_project_id ≠ target_project_id`) and skip the branch switch; extract `activate_mr`, `select_fork_mr`, and `select_local_mr` helpers for clarity.
- **`lua/gitlab/hunks.lua`** —  pass `head_sha` to `parse_hunks_and_diff`. The function now accepts and forwards head_sha to the git diff command, ensuring the diff is computed between the correct base and head SHAs of the merge request rather than comparing against the working tree. This make diff working with the new strategy

### Suggestion

Probably it will be a good idea to use this approach also for local merge request and avoid the fetching of branch and checkout to it. This approach has several benefits

 - No branch switch needed — your working tree stays untouched
 - No clean tree requirement — you can have uncommitted changes
 - Faster — no git switch, no buffer reloads
 - Review any MR from any branch — you don't need the branch locally at all

### DIFF
--- a/cmd/app/git/git.go
+++ b/cmd/app/git/git.go
@@ -70,7 +70,7 @@ func NewGitData(remote string, gitlabUrl string, g GitManager) (GitData, error) 
 	// remove part of the hostname from the parsed namespace
 	url_re := regexp.MustCompile(`[^\/]\/([^\/].*)$`)
 	url_matches := url_re.FindStringSubmatch(gitlabUrl)
-	var namespace string = matches[1]
+	namespace := matches[1]
 	if len(url_matches) == 2 {
 		namespace = strings.TrimLeft(strings.TrimPrefix(namespace, url_matches[1]), "/")
 	}
--- a/cmd/app/git/git.go
+++ b/cmd/app/git/git.go
@@ -12,6 +12,8 @@ type GitManager interface {
 	GetProjectUrlFromNativeGitCmd(remote string) (url string, err error)
 	GetCurrentBranchNameFromNativeGitCmd() (string, error)
 	GetLatestCommitOnRemote(remote string, branchName string) (string, error)
+	GetMRHeadCommit(mrIID int64) (string, error)
+	FetchMRHead(remote string, mrIID int64) error
 }
 
 type GitData struct {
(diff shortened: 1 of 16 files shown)


jakubbortlik:
Hi @Clement-Hue, I'm glad to see more contributors to the plugin. Just a note regarding your suggestion to use your new approach also for local merge requests. The benefit of using the local files for a review are that you can use all the benefits of LSP, code navigation, linting, etc. As for the perceived benefits of your approach, you can achieve at least some of them by using [git worktrees](https://git-scm.com/docs/git-worktree). With worktrees you can have several versions of the same repo checked out at the same time and you don't worry about switching branches, stashing changes, and a clean tree. I find this a superior workflow for multiple feature branches regardless of merge requests.

The only reason I'm missing the possibility to review a MR without a local branch is for a closed MR where the feature branch has already been deleted.

I think it would be great if gitlab.nvim could support both these workflows.

jakubbortlik:
By the way, @Clement-Hue, the [pipeline has failed](https://github.com/harrisoncramer/gitlab.nvim/actions/runs/24899727133/job/73025710717?pr=546), so please have a look and fix the issue.

Clement-Hue:
I fixed the lint error. You are right we are loosing the lsp features. For the moment I use it without, the gitlab plugin in vscode also does not support lsp feature since as you said  there is nothing local. It would be great indeed to have this possibility, it is nonetheless usable.

cmd/app/git/git.go
```diff
@@ -125,6 +127,40 @@ func (g Git) RefreshProjectInfo(remote string) error {
 	return nil
 }
 
+/* Fetches the head ref of a merge request so that fork MR commits are available locally */
+func (g Git) FetchMRHead(remote string, mrIID int64) error {
+	ref := fmt.Sprintf("refs/merge-requests/%d/head", mrIID)
+	localRef := ref
+	cmd := exec.Command("git", "fetch", remote, fmt.Sprintf("%s:%s", ref, localRef))
```
jakubbortlik: What's the point of using two variables for the same value?

lua/gitlab/actions/merge_requests.lua
```diff
@@ -4,6 +4,54 @@ local git = require("gitlab.git")
...
+    end)
+  end)
+end
+
+---Selects a fork MR by restarting the server without switching branches.
+---@param choice table
+---@param opts ChooseMergeRequestOptions
+local select_fork_mr = function(choice, opts)
+  activate_mr(choice, opts)
+end
```
jakubbortlik, line 26: You don't need this empty wrapper function, do you?

Clement-Hue: yes it was just to be consistent with the select_local_mr in naming

PR #548, opened 2026-04-24
fix(build): allow force-rebuilding server binary

### LINKED ISSUE #545: Doesn't build binary
## Prerequsities

## Bug Description

The binary is not rebuilt after running gitlab.server.build(). The problem is that:
1. The function is first run [during the setup](https://github.com/harrisoncramer/gitlab.nvim/blob/956d21a095cbe7d4338d0df09ac6c88f2866e2d9/lua/gitlab/init.lua#L49)
2. In the initial `build` run, the [state.settings.server.binary](https://github.com/harrisoncramer/gitlab.nvim/blob/4a890ea41c7e05da6456890a7ad704bad5804685/lua/gitlab/server.lua#L151) field is set
3. If I run `gitlab.server.build(true)` again to force-rebuild the server, the function doesn't get past the [check if the server exists](https://github.com/harrisoncramer/gitlab.nvim/blob/4a890ea41c7e05da6456890a7ad704bad5804685/lua/gitlab/server.lua#L125-L127) and returns early never rebuilding the server.
4. I have to delete the binary in order to rebuild the server

## Reproduction Steps

1. Check out the feature branch
2. Open Neovim
3. run `lua require("gitlab.server").build(true) -- should force-rebuild the server and print "Installed successfully!" but that doesn't happen

harrisoncramer:
## Summary

- Fixes the `build(true)` force-rebuild not working after the initial build
- The `server.binary` field was being set during the first build, causing subsequent calls to think it was a user-provided path and skip rebuilding
- Now tracks user-provided binaries via a `binary_provided` flag set during config merge

Closes #545

### DIFF
--- a/lua/gitlab/server.lua
+++ b/lua/gitlab/server.lua
@@ -122,8 +122,7 @@ M.build = function(override)
   state.settings.root_path = u.get_root_path()
 
   -- If the user provided a path to the server, don't build it.
-  if state.settings.server.binary ~= nil then
-    state.settings.server.binary_provided = true
+  if state.settings.server.binary_provided then
     local binary_exists = vim.loop.fs_stat(state.settings.server.binary)
     if binary_exists == nil then
       u.notify(
--- a/lua/gitlab/state.lua
+++ b/lua/gitlab/state.lua
@@ -454,6 +454,9 @@ end
 ---@param args Settings
 ---@return Settings
 M.merge_settings = function(args)
+  if args.server and args.server.binary ~= nil then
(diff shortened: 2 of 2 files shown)


harrisoncramer:
@jakubbortlik does this solve your problem in https://github.com/harrisoncramer/gitlab.nvim/issues/545

jakubbortlik:
> @jakubbortlik does this solve your problem in #545

Hi Harrison, yes, this fixes the problem. Thanks!

PR #549, opened 2026-04-24
feat: Fetch all merge request beyond the default 20

Clement-Hue:
**Summary**

Fetch all project merge requests from the GitLab API by paginating through results server-side, instead of relying on a single page (default 20). This ensures the merge request chooser shows the complete list regardless of project size.

**Problem**

The `ListProjectMergeRequests` call only returned the first page of results (default 20 MRs per the GitLab API). Projects with more merge requests had entries silently truncated, so users could not select older or less recent MRs from the picker. An earlier attempt added a `per_page` option on the Lua side, but that only shifted the limit instead of solving the truncation.

**Changes**

- `cmd/app/merge_requests.go`: loop over all pages using `res.NextPage`, accumulating results into a single `mergeRequests` slice before responding.
- Default `PerPage` to 100, to minimize round-trips.

### DIFF
--- a/cmd/app/merge_requests.go
+++ b/cmd/app/merge_requests.go
@@ -35,16 +35,23 @@ func (a mergeRequestListerService) ServeHTTP(w http.ResponseWriter, r *http.Requ
 		payload.Scope = gitlab.Ptr("all")
 	}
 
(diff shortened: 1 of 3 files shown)


harrisoncramer:
Thank you!

PR #554, opened 2026-04-29
fix: encode empty merge_requests body as JSON object instead of array

Cliffback:
When choose_merge_request() is called without arguments, the body function returns an empty Lua table {} which vim.json.encode serializes as a JSON array []. The Go server expects a JSON object and fails with: 'cannot unmarshal array into Go value of type ListProjectMergeRequestsOptions'

Using vim.empty_dict() ensures the empty table is encoded as {} instead.


jakubbortlik:
Thanks for reporting and fixing this!

Cliffback:
No problem, happy to help!

PR #557, opened 2026-05-05
fix(git): handle detached HEAD silently

### LINKED ISSUE #555: choose_merge_request crashes on detached HEAD
## Prerequsities

## Setup Configuration and Environment

Default configuration; no environment variables relevant to this path.

## Bug Description

`choose_merge_request()` crashes when the working tree is in detached
HEAD (or in a fresh repo without commits):

```
.../lua/gitlab/git.lua:112: attempt to concatenate local 'err' (a nil value)
stack traceback:
  .../lua/gitlab/git.lua:112: in function 'get_current_branch'
  .../lua/gitlab/actions/merge_requests.lua:33: in function 'cb'
```

`git branch --show-current` returns an empty string with exit code 0
in that case, so `err` is `nil` and the error branch in `git.lua:112`
concatenates `nil` into the notify message.

The only caller of `get_current_branch()` in this flow is
`merge_requests.lua:33`, which uses the return value to skip a
clean-tree check and a no-op `switch_branch` when the user is already
on the source branch, so detached HEAD has no semantic meaning here
beyond the crash itself.

Happy to open a PR.

## Reproduction Steps

1. Check out an arbitrary commit in detached HEAD mode
2. In Neovim: `:lua require("gitlab").choose_merge_request()`
3. Pick any MR → crash

seflue:
Closes #555.

`git branch --show-current` returns an empty string with exit code 0 when HEAD is detached or the repo has no commits. That's a valid state, not an error. The old code routed it through the error branch anyway, so the operation was cancelled with an error message instead of proceeding.

The fix splits the two cases. In detached HEAD, the picker now proceeds with the check-out (the caller in `merge_requests.lua:33` already handles nil). Real git failures still notify ERROR.

### DIFF
--- a/lua/gitlab/git.lua
+++ b/lua/gitlab/git.lua
@@ -104,12 +104,16 @@ M.get_ahead_behind = function(current_branch, remote_branch)
   return tonumber(ahead), tonumber(behind)
 end
 
----Return the name of the current branch or nil if it can't be retrieved
+---Return the name of the current branch, or nil (detached HEAD or git failure).
 ---@return string|nil
 M.get_current_branch = function()
   local current_branch, err = run_system({ "git", "branch", "--show-current" })
-  if err or current_branch == "" then
+  if err then
     require("gitlab.utils").notify("Could not get current branch: " .. err, vim.log.levels.ERROR)
+    return nil
+  end
+  if current_branch == "" then
+    -- detached HEAD: `git branch --show-current` returns empty + exit 0
     return nil
   end
   return current_branch


lua/gitlab/git.lua
```diff
@@ -104,14 +104,18 @@ M.get_ahead_behind = function(current_branch, remote_branch)
...
 ---@return string|nil
 M.get_current_branch = function()
   local current_branch, err = run_system({ "git", "branch", "--show-current" })
-  if err or current_branch == "" then
+  if err then
     require("gitlab.utils").notify("Could not get current branch: " .. err, vim.log.levels.ERROR)
     return nil
   end
+  if current_branch == "" then
+    -- detached HEAD: `git branch --show-current` returns empty + exit 0
```
jakubbortlik, line 116: I'm thinking if the commit the head is detached at is not useful information that we could notify the user about. Maybe they'd like to check it out again later and they would find the notification useful.

seflue: I'm not sure, if I understand, what you mean. The intent of the function as I understood it was "give me the current branch, so I can decide if it is necessary to actually check the MR branch out or not". If there is no branch checked out, I expect the plugin to do exactly that.

A use case (and that is how I actually ran into that problem): Creating a work tree for review, check it out in detached head mode on the branch I currently am (e.g. main because I don't want to search for any branch - the plugin actually does that step for me anyway - and I cannot check out the same branch in two work trees). Then from this work tree start neovim and start the review. I don't want the plugin stand in my way with assumptions on current branches or whatnot, it should just focus on it's purpose with as least friction as possible: let me review the desired merge request.

jakubbortlik: I see. I only get into detached head when I want to inspect some previous state of the code base and in that case I think I would like to be notified about the state changing. But your use case makes sense.

jakubbortlik:
Hi @seflue, thanks for the contribution.

PR #572, opened 2026-07-01
fix: only check Go version when building the server

snejugal:
I use NixOS and prefer to keep my environment clean. As such, I prebuilt the plugin's server binary using Nix and specified its path in `setup`. However, `setup` forcibly checks Go version and aborts if I don't have it installed, which doesn't make sense because Go is only required to build the server and I already have one. Thus I moved the check to `server.build` function when it actually needs to build the server. I also touched `version.lua` to remove the annoying `go: command not found` message when starting Neovim (I'm not sure if this is a good way to fix it since I'm not a Lua programmer, so I'm open to suggestions).

### DIFF
--- a/lua/gitlab/server.lua
+++ b/lua/gitlab/server.lua
@@ -4,6 +4,7 @@
 local state = require("gitlab.state")
 local u = require("gitlab.utils")
 local job = require("gitlab.job")
+local version = require("gitlab.version")
 local M = {}
 
 -- Builds the binary if it doesn't exist, and starts the server. If the pre-existing binary has an older
@@ -23,8 +24,8 @@ M.build_and_start = function(callback)
       callback()
       return
     end
-    M.get_version(function(version)
-      if version.plugin_version ~= version.binary_version then
+    M.get_version(function(versions)
+      if versions.plugin_version ~= versions.binary_version then
         M.shutdown(function()
           if M.build(true) then
             M.start(callback)
@@ -116,6 +117,12 @@ M.build = function(override)
     return
   end
 
+  local version_issue = version.check_go_version()
+  if version_issue ~= nil then
+    u.notify(version_issue, vim.log.levels.ERROR)
+    return
+  end
+
   -- If the user did not provide a path, we build it and place it in either the data path, or the
   -- first writable path we find in the runtime.
   local datapath = vim.fn.stdpath("data")
@@ -150,9 +157,9 @@ M.build = function(override)
   local version_output = vim
     .system({ "git", "describe", "--tags", "--always" }, { cwd = state.settings.root_path })
     :wait()
-  local version = version_output.code == 0 and vim.trim(version_output.stdout) or "unknown"
+  local plugin_version = version_output.code == 0 and vim.trim(version_output.stdout) or "unknown"
 
-  local ldflags = string.format("-X main.Version=%s", version)
+  local ldflags = string.format("-X main.Version=%s", plugin_version)
   local res = vim
     .system(
       { "go", "build", "-buildvcs=false", "-ldflags", ldflags, "-o", state.settings.server.binary },
--- a/lua/gitlab/server.lua
+++ b/lua/gitlab/server.lua
@@ -106,20 +106,11 @@ M.build = function(override)
 
   -- If the user provided a path to the server, don't build it.
   if state.settings.server.binary_provided then
-    local binary_exists = vim.loop.fs_stat(state.settings.server.binary)
-    if binary_exists == nil then
-      u.notify(
-        string.format("The user-provided server path (%s) does not exist.", state.settings.server.binary),
-        vim.log.levels.ERROR
-      )
-      return false
-    end
     return
   end
 
   local version_issue = version.check_go_version()
   if version_issue ~= nil then
-    u.notify(version_issue, vim.log.levels.ERROR)
     return
   end
 
--- a/lua/gitlab/version.lua
+++ b/lua/gitlab/version.lua
@@ -1,7 +1,7 @@
 local M = {}
 
 M.is_go_valid = function()
-  local go_version = io.popen("go version"):read("*a")
+  local go_version = io.popen("go version 2>&1"):read("*a")
   if go_version then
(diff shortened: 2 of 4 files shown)


lua/gitlab/version.lua
```diff
@@ -1,7 +1,7 @@
 local M = {}
 
 M.is_go_valid = function()
-  local go_version = io.popen("go version"):read("*a")
+  local go_version = io.popen("go version 2>&1"):read("*a")
```
jakubbortlik: Hi @snejugal, maybe instead you could modify the health check (health.lua module) to only call check_go_version when the state.settings.server.binary_provided option is nil.

snejugal: I can see that asking the user to install Go is unnecessary when they provide a prebuilt binary. Still, if the user is missing both of those, this function will still be called, and non-captured `stderr` causes the editor to misrender slightly. Compare:

<details>
<summary><code>:checkhealth</code> when <code>stderr</code> is not captured</summary>

https://github.com/user-attachments/assets/bb756a84-a22a-4918-9f36-0de70936db82

</details>

<details>
<summary><code>:checkhealth</code> when <code>stderr</code> is captured</summary>

https://github.com/user-attachments/assets/a5490cb8-93a6-431a-b917-597197f14a7f

</details>

so I believe a patch to capture `stderr` here is still needed.

jakubbortlik: I see, maybe we could use vim.system instead od io.popen, as we do in most other places. I'll have a look next week when I'm back from vacation.

snejugal: Just tried to tinker with `vim.system` myself, definitely seems to be a better option since no shell is spawned and no `command not found` message is produced in the first place. Pushed https://github.com/harrisoncramer/gitlab.nvim/pull/572/changes/d5fa4d460e77a43406aa826e6a04be31c5dc39a6, please take a look when you have the time.

lua/gitlab/version.lua
```diff
@@ -1,14 +1,15 @@
...
-      return false
-    end
+  local has_go, go = pcall(vim.system, {"go", "version"});
+  if not has_go then
+    return false
+  end
+
+  local go_version = go:wait().stdout;
+  local major, minor, _ = go_version:match("(%d+)%.(%d+)%.?(%d*)")
+  if major and tonumber(major) >= 1 and tonumber(minor) >= 25 then
```
jakubbortlik: I've just remembered, there's a more robust way to compare versions in Neovim: https://neovim.io/doc/user/lua/#vim.version. Could you please try and replace the current way?

snejugal: Done, see https://github.com/harrisoncramer/gitlab.nvim/pull/572/commits/f3771da0568cbd86cfc181e155dabb71f222d01a

lua/gitlab/version.lua
```diff
@@ -1,17 +1,13 @@
...
-      return false
-    end
-  else
+  local has_go, go = pcall(vim.system, { "go", "version" })
+  if not has_go then
     return false
   end
+
+  local go_version = vim.version.parse(go:wait().stdout, { strict = false })
+  return vim.version.ge(go_version, { 1, 25, 0 })
```
jakubbortlik: Hi @snejugal, I'm finally having a better look at your PR.

`vim.version.parse` can return `nil`, but `vim.version.ge` expects type `string|number[]|vim.Version`. Can you add a "nil guard" before you call `vim.version.ge`?

lua/gitlab/server.lua
jakubbortlik: I think it would be nice to move this check to the health.lua module and at the same time improve the [Go version check](https://github.com/snejugal/gitlab.nvim/blob/1f7c9e50b5c48aa67e35b0e81d70994fea0347dd/lua/gitlab/health.lua#L50-L53).

My idea is that in server.lua we simply return `if state.settings.server.binary_provided` is true. In health.lua, we'd first check if `state.settings.server.binary_provided` was true, if it was, we'd perform the `binary_exists` check now located in server.lua - it that failed, the error wouldn't be logged directly as it is now, but it would be added to the `errors` list in health.lua - that would automatically warn the user when initializing the plugin or when running `:checkhealth gitlab` manually. If `state.settings.server.binary_provided` was false, then we'd check the Go version and if there was a problem with go, this would again be added to the `errors` table. Now it's in `warnings` which doesn't make sense - either the user doesn't need Go at all, then a problem with Go should not surface in the health check, or the user needs Go and then it's a blocking error.

snejugal: I assume that when Go is missing, the error shouldn't be logged directly as well. Done in https://github.com/harrisoncramer/gitlab.nvim/pull/572/commits/c7857766c105a2dcfba6fb2e80124207e2f8e1da, here's how it looks:

image

image

image

lua/gitlab/version.lua
snejugal: Done in https://github.com/harrisoncramer/gitlab.nvim/pull/572/commits/ee4eca1a1658b7c59449b01e428ebf57871a6359

lua/gitlab/version.lua
```diff
@@ -1,17 +1,13 @@
...
-      return false
-    end
-  else
+  local has_go, go = pcall(vim.system, { "go", "version" })
+  if not has_go then
     return false
   end
+
+  local go_version = vim.version.parse(go:wait().stdout, { strict = false })
+  return go_version ~= nil and vim.version.ge(go_version, { 1, 25, 0 })
```
jakubbortlik: @snejugal Could you please update the version to 1.25.1 - this is the actual version in
`go.mod`. Ideally, I'd like to see the version pulled out to a variable and used in the return value of check_go_version as well, so that they don't drift.

snejugal: Done, see https://github.com/harrisoncramer/gitlab.nvim/pull/572/commits/46a49be1f90a1fe5f8184d2bfbe0b4d15966a1d8

jakubbortlik:
@snejugal thanks for the contribution and collaboration.

PR #573, opened 2026-07-11
fix: set diffview hook for added files

jakubbortlik:
The hook for setting up keymaps and autocommands didn't fire correctly if the first file in the review was a new file, so it was diffed against diffview://null, which caused some autocommands to not fire appropriately or some state (reviewer.tabid) not to be set correctly.


lua/gitlab/actions/discussions/init.lua
```diff
@@ -176,10 +177,6 @@ M.open = function(callback, view_type)
   if type(callback) == "function" then
     callback()
   end
-
-  vim.schedule(function()
-    M.refresh_diagnostics()
-  end)
```
jakubbortlik, line 182: This diagnostics refresh can cause errors when the reviewer is not available yet. It is not even necessary, since diagnostics are refreshed through `draft_notes.rebuild_view` on line 175.

Issue #463, opened 2025-01-23
Show MR Activity in a new "tab"


jakubbortlik:
## Feature Description
The Gitlab Overview page shows the "Activity" including MR renames, commits, review requests, approvals, etc. An example:

![Image](https://github.com/user-attachments/assets/60a4f5a7-e1b7-4c36-8544-a72b74b4bf3b)

I'd find it helpful if this information was also shown in `gitlab.nvim`. I imagine it could be another "tab" next to "Comments" and "Notes".

I think this should not even be that difficult to prepare a simple listing of the items. I can imagine it would be useful to make it interactive by e.g., enabling to easily show the commits or diffs just like Gitlab makes the commit hashes clickable and adds a "Compare with previous version" button. This might even be useful as a workaround for #450.

PR #577, opened 2026-07-29
feat(reviewer): browse MR commit history

### LINKED ISSUE #450: Navigate through MR by commit
## Feature Description

Gitlab MRs currently allow for reviewing changes made on each commit and commenting on specific SHAs. Providing a way of focusing on a specific commit could help with reviews that involve well-constructed incremental changes.

### LINKED ISSUE #576: Commit-by-commit review of a merge request
Reviewing a large MR as a single combined diff loses the commit-by-commit story: you can't see how the change was built up as you read it.

I'd like to browse an MR one commit at a time, the way you'd read a branch's history.

I have this implemented and can open a PR.

seflue:
Closes #576.

On larger MRs I like to review commit by commit. My teammates put real effort into structuring their history, and getting value out of that means the tooling has to support reviewing it that way.

So I added a commit browser: `glh` opens the MR's history so it can be read one commit at a time.

I also added commenting from that view: `c` comments on the line under the cursor while browsing. A commit shows its own `commit^..commit` diff while GitLab positions comments against the MR's base..head, so a line is only commentable when it maps cleanly onto base..head; when a later commit rewrote it, `]v` / `[v` move to a version where it does. I went with "comment where it maps, navigate otherwise", but I'm open to other takes.

The regular reviewer is untouched. Line math comes from `git diff` and `git log -L` against the MR's own SHAs, so it ignores the working tree, and the parsers have unit tests.

One dependency outside the browser: `refresh_diagnostics()` now returns quietly when no regular reviewer view is open. Commenting routes through the shared comment path, which rebuilds the views; with no reviewer open that previously errored on an otherwise successful comment.

### DIFF
--- a/after/syntax/gitlab.vim
+++ b/after/syntax/gitlab.vim
@@ -11,10 +11,13 @@ let formatted_date = '\w\+ \{1,2}\d\{1,2}, \d\{4}'
 let absolute_time = '\d\{2}/\d\{2}/\d\{4} at \d\{2}:\d\{2}'
 let date = '\%(' . time_ago . '\|' . formatted_date . '\|' . absolute_time . '\|just now\)'
 
-let published = date . ' \%(' . g:gitlab_discussion_tree_resolved . '\|' . g:gitlab_discussion_tree_unresolved . '\|' . g:gitlab_discussion_tree_unlinked . '\)\?'
+" Commits are referenced by the first 7 characters of their SHA, e.g. '1a2b3c4'
+let commit_ref = '[0-9a-f]\{7}'
+
+let published = date . '\%( ' . commit_ref . '\)\?' . ' \%(' . g:gitlab_discussion_tree_resolved . '\|' . g:gitlab_discussion_tree_unresolved . '\|' . g:gitlab_discussion_tree_unlinked . '\)\?'
 let state = ' \%(' . published . '\|' . g:gitlab_discussion_tree_draft . '\)'
 
-execute 'syntax match GitlabNoteHeader "' . expanders . username . state . '" contains=GitlabDate,GitlabUnresolved,GitlabUnlinked,GitlabResolved,GitlabExpander,GitlabDraft,GitlabUsername'
+execute 'syntax match GitlabNoteHeader "' . expanders . username . state . '" contains=GitlabDate,GitlabUnresolved,GitlabUnlinked,GitlabResolved,GitlabExpander,GitlabDraft,GitlabUsername,GitlabCommit'
 
 execute 'syntax match GitlabDate "' . date . '" contained'
 execute 'syntax match GitlabUnresolved "' . g:gitlab_discussion_tree_unresolved . '" contained'
@@ -24,5 +27,6 @@ execute 'syntax match GitlabExpander "' . expanders . '" contained'
 execute 'syntax match GitlabDraft "' . g:gitlab_discussion_tree_draft . '" contained'
 execute 'syntax match GitlabUsername "' . username . '" contained'
 execute 'syntax match GitlabMention "' . username . '"'
+execute 'syntax match GitlabCommit "' . commit_ref . '" contained'
 
 let b:current_syntax = 'gitlab'
--- a/lua/gitlab/reviewer/history.lua
+++ b/lua/gitlab/reviewer/history.lua
@@ -8,10 +8,12 @@
 -- start_sha on its first parent and head_sha on the commit, plus a top-level commit_id;
 -- anything else is rejected with "commit_id does not match the diff refs".
 --
--- GitLab stores base_sha/start_sha as the MR base whatever we send. An old-side (left
--- window) line, one the commit deletes, is then numbered against a diff other than the one
--- the browser shows, with no verified anchor, so we refuse to comment there (see
--- create_comment).
+-- GitLab stores base_sha/start_sha as the MR base whatever we send, and renumbers the
+-- position's old_line to match, so the position carries the commit's own line numbers and
+-- GitLab does the translation (measured, see
+-- memories/gitlab-api-experiments/replay-percommit-context.sh). The old (left window) side
+-- stays refused: a line the commit deletes need not exist in the MR base at all, and that
+-- case is unmeasured (see create_comment).
 --
 -- Those commit-anchored notes are marked in the browser and nowhere else, since their
 -- lines only mean anything in the commit's own diff (see indicators/common.lua).
@@ -19,6 +21,8 @@
 local List = require("gitlab.utils.list")
 local u = require("gitlab.utils")
 local state = require("gitlab.state")
+local hunks = require("gitlab.hunks")
+local Location = require("gitlab.reviewer.location")
 local reviewer = require("gitlab.reviewer")
 local async = require("diffview.async")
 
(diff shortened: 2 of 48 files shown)


jakubbortlik:
Hi Sebastian, I think this will be a welcome improvement to the plugin (I believe there was already the issue #450 that tracked the feature request, so I guess you can close your new issue #576). However, I don't know whether Harrison, the repo owner, has the time and motivation to review any PRs right now. I certainly want to have a look at it some time, but at the moment my priority is fixing #386 which has been around for almost two years and I believe the fix will also influence the your PR - I would like to radically simplify and fix the way how the Location data are calculated for normal comments, as currently the plugin does too much work (like parsing diff hunks three times instead of once) and it doesn't even produce the correct data (line codes, modification types, and line numbers in ranged comments).

I've just briefly had a look at your PR and I have some general suggestions. I believe it would be more user-friendly if the commenting worked similar to how comments can be created in the standard reviewer and how comments on individual commits can be created in Gitlab online: Users should be able to create ranged comments on any commit, irrespective of whether the code was changed later, otherwise I don't see the point of commenting on individual commits if I can't really comment on them. If it's currently impossible to create comments on lines that are changed in later commits, then we should find out how to make it possible instead of forcing the user to comment on a different commit. I suggest you investigate what payload Gitlab creates when creating commit-bound comments in the browser - I've briefly had a look and it sends different "base_sha", "head_sha", "start_sha" values, and I think the way to go will be to also send the [correct SHAs](https://github.com/harrisoncramer/gitlab.nvim/blob/31aba9d65f10d3f13e902ed9bcad593b002c9955/lua/gitlab/actions/comment.lua#L79-L81).

I would like to fix #386 as soon as possible and then I will be open to reviewing your PR but I can't promise anything.

lua/gitlab/reviewer/history.lua
```diff
@@ -0,0 +1,404 @@
...
+  end
+
+  if keymaps.reviewer.create_comment ~= false then
+    vim.keymap.set("n", keymaps.reviewer.create_comment, function()
+      require("gitlab").history_create_comment()
+    end, { buffer = bufnr, desc = "Comment on this line (commit browser)" })
+  end
+
+  if keymaps.reviewer.browse_next_version ~= false then
+    vim.keymap.set("n", keymaps.reviewer.browse_next_version, function()
```
jakubbortlik: This causes the plugin to fail opening the commit history, if it doesn't fail for you, you probably have `keymaps.reviewer.browse_next_version` (and `browse_prev_version`) mapped in your personal config:
```suggestion
  if keymaps.reviewer.history_next_version ~= false then
    vim.keymap.set("n", keymaps.reviewer.history_next_version, function()
```

seflue:
@jakubbortlik  thanks for raising this. Going by the docs, the discussions position is always built from the MR's diff refs, so I assumed comments could only target the MR head. I just ran some tests against a live instance and got it working. It's the same endpoint we already use. With the position's `head_sha` pointing at the commit and a `commit_id` alongside, GitLab interprets it as a comment on an intermediate commit.

I'd still like to keep two things. Walking through a file's versions is useful on its own, independent of commenting, so I'd leave it in as a browsing feature. And when a line commented on an earlier commit gets overwritten later in the same MR, the user should notice. GitLab already marks such comments outdated, and I'd rather show it than hide the case.

As for #386, I'll keep going in the meantime and rebase once it's merged.

lua/gitlab/reviewer/history.lua
seflue: Thanks for catching this. Fixed and force-pushed, with a regression test that flags this kind of bug.

seflue:
I continued to work on the implementation. This is what I have achieved now:

- commits can be annotated with comments, draft mode works
- commit comments are marked with the commit hash in the discussion tree
- the discussion window has been integrated into the commit-history tab
- navigation between comments and code works consistently and switches to the expected tab

To show the discussion window in both tabs (reviewer tab and commit-history tab) I had to add a window registry. There used to be exactly one discussion window: its window id, its two buffers and the current view type were single fields in `discussions/init.lua`, and opening the tree in a second tab overwrote all of them. The first window stayed on screen with nothing pointing at it. The registry now holds one entry per tab. Buffers and trees stay shared, so both tabs show the same content; only the window, its view type and the cursor position belong to the tab. Feedback on this one matters most to me, since everything else sits on top of it.

I did one refactoring to make the code easier to read. It was not necessary for the implementation, but it helped me understand it. `move_to_discussion_tree` had grown three nested closures and the tab id had to reach the innermost one, so I pulled it apart into two named functions and put tests around it. I would happily do more of that in this area if you want it, and I can just as well drop this one if you prefer a narrow diff.

I also fixed a small thing: jumping from a line without a comment warned about missing diagnostics, which confused me. It now says there is no comment on this line.

The last point is an idea, not implemented yet. Commit comments and MR comments currently live in the same list, told apart only by the hash. I am thinking about separating them with a third winbar tab for the commit-specific comments. Typing `c` would then cycle through three views instead of toggling two. Does a third tab fit how you see the discussion window, or would you rather keep two? I also thought about making both variants available through configuration, but honestly I think three winbar tabs are more consistent with the overall design.

The code sits on top of #581.

jakubbortlik:
Hi Sebastian. Please have a look at #582 - I'd be grateful if you could do a genuine review and also if you told me if that PR will be useful for your current PR or I could do something differently to make your work on this easier.

after/syntax/gitlab.vim
jakubbortlik, line 1: I'm working on a re-implementation (#584) of the discussion tree to use virtual text for indentation instead of literal spaces and as a part of that I would like to change the syntax highlighting of the discussion tree header nodes to extmarks so this file will almost completely be gone. But adding the commit_ref will be very easy in the new approach.

jakubbortlik:
Hi Sebastian. I'm having a look at your PR and I definitely want to make the feature part of the plugin, but the PR is huge and I would like to propose some changes to limit the scope, so that it becomes manageable. I'd like to also keep working on some other fixes/features in the meantime :) I realize that a large part of the additions are tests, which is great, but tests are code too and need to be reviewed and maintained, so this doesn't make the review any simpler :)

Here are some suggestions to make this more likely to get merged. I hope these comments make sense and will be actionable.

## Split the PR
Please pull out commits not directly related to the this feature into separate PRs if possible (mainly jumping to the discussion). Consider if other parts (e.g., the window registry and allowing the discussion split in several tabs) could also be pulled out.

## Consider upstreaming some features to diffview+
The commit navigation feature: `]v` and `[v` sounds like something that could be useful in `DiffviewFileHistory` in general. Would it be possible to make it part of `diffview+` instead? The way I understand it "is navigation only, and independent of commenting" so it should be possible.

## Commenting on commits
I need to understand the use case or your intended workflow better, as I don't use the commit view in Gitlab too often (but that might change if it's available in `gitlab.nvim` :grin:). As far as I know the web UI makes it possible to comment on a commit diff the same way as on a MR diff - irrespective if the line(s) change in a later version. I feel like we should also allow this and I don't understand why the comment should be instead made on some other commit that touches that line (see "Outdated comments" below) or why commenting on the old version is not possible in this PR (I admit I didn't try too hard to understand the commit message :grin:).

I believe it should be possible to reuse the existing way of getting the location data - modify `reviewer.get_reviewer_data()` to allow getting `ReviewerData` from the `DiffviewFileHistory` rather than from the main `Diffview` and then feed it to `hunks.get_hunks()` and to `Location.new()`. I would strongly prefer if `history.lua` and `history_diff.lua` didn't implement their own context-getting, location-building, and diff-parsing logic unless it's absolutely necessary.

### Outdated comments
You wrote before that you'd like to keep this behaviour:
> when a line commented on an earlier commit gets overwritten later in the same MR, the user should notice. GitLab already marks such comments outdated, and I'd rather show it than hide the case.

Does it mean A) the user should notice when creating the comment or B) we should somehow mark such comments in the discussion tree?

Regarding A: I agree that a notification would be useful, maybe there could be two kinds of commenting keybindings: one to comment on this exact commit and another one to comment on the last commit that touches this line (I believe this is closer to how you intended the commenting to work).

Regarding B: Outdated comments often have incorrectly placed diagnostics, which is something we should fix regardless of whether the comment was on a specific commit or just in an older version of the MR (comment in the regular reviewer). Maybe we could also implement something like "jump to the diff version on which this comment was originally created" which is conceptually similar to your feature, except for the distinction between "versions" and "commits" - comparing the various versions of v MR would be a useful feature that could be a followup for this PR.

Just FYI, my long-lived draft #504 enables something similar - to edit a comment with a preview diff in a new tab that uses the [original head sha](https://github.com/jakubbortlik/gitlab.nvim/blob/75e70fe01ad04dd2928d1038881a8bc3e794a300/lua/gitlab/actions/suggestions.lua#L418-L419), but that's a workaround, not a proper solution for the use case and I'm not particularly pressed to finish that PR since I currently prioritize fixing bugs and use that suggestion-preview branch locally for my own daily work...

# Other comments

## Bugs
1. I've tried your branch and when I open the regular reviewer with the discussion tree and then run `glh` I see the discussion tree twice in the commit view tab:
image

2. The spinner in the two tabs is broken because both winbars update the same [spinner_index](https://github.com/seflue/gitlab.nvim/blob/86bf3bcbf7622dadb7bad3aa4cdbcecf3b27c67a/lua/gitlab/actions/discussions/winbar.lua#L60) - that's not a bug you introduced, it's part of a faulty implementation on my side and I'm working on a fix so you don't need to worry about it (a fix definitely should not be part of this PR :grin:).

## Separate winbar tab
> Commit comments and MR comments currently live in the same list, told apart only by the hash. I am thinking about separating them with a third winbar tab for the commit-specific comments.

In #463 I proposed something related - use a third tab to show the MR activity - I miss this kind of information in `gitlab.nvim` quite a lot and that's also information that is present in most MRs. In contrast, unlinked notes are, in my experience, rather infrequent and I imagine for some users so will be comments on specific commits. I'm not really convinced they need their own tabs at all and would actually consider putting them all into a single list, differentiated by e.g. the commit hash or a suitable icon. I use a right-hand side vertical split for the discussion tree and I find the winbar already rather crammed. What do you think?

jakubbortlik:
Hi Sebastian, I've opened #586 to fix the spinner issues I mentioned earlier:
>     2. The spinner in the two tabs is broken because both winbars update the same [spinner_index](https://github.com/seflue/gitlab.nvim/blob/86bf3bcbf7622dadb7bad3aa4cdbcecf3b27c67a/lua/gitlab/actions/discussions/winbar.lua#L60) - that's not a bug you introduced, it's part of a faulty implementation on my side and I'm working on a fix so you don't need to worry about it (a fix definitely should not be part of this PR 😁).

It will probably conflict with your changes to the winbar, but I hope it should be easy to resolve.

PR #580, opened 2026-08-02
fix: hunk header counts and the deleted-line check

seflue:
Git omits the `,N` in a hunk header when N is 1, so `@@ -5 +4,0 @@` is a one-line deletion. `parse_possible_hunk_headers` reads the empty capture as 0, which makes that header look like `@@ -5,0 +4,0 @@`: `is_range_zero` short-circuits `get_modification_type_from_new_sha`, and the unmodified line above the deletion comes back as `added`. The same deletion over two lines classifies correctly. The first commit reads the empty capture as 1.

Found this while working on #577.

While I was in there I replaced the body walk in `line_was_removed` with a range check: with `--unified=0` a hunk has no context lines, so the old range already answers the question. No behaviour change. The old side had no tests, so that commit adds some.

Putting `get_modification_type` under test properly is a different matter; I will write that up in #386. Related to that issue, not a fix for it.

### DIFF
--- a/lua/gitlab/hunks.lua
+++ b/lua/gitlab/hunks.lua
@@ -25,11 +25,13 @@ M.parse_possible_hunk_headers = function(line)
     --  @@ -41,0 +42,4 @@ ...
     local old_start, old_range, new_start, new_range = line:match("@@+ %-(%d+),?(%d*) %+(%d+),?(%d*) @@+")
 
+    -- Git omits the ",N" count when it is exactly 1, so an empty capture means 1,
+    -- while a captured "0" means a genuine zero-length range (pure insertion/deletion).
     return {
       old_line = tonumber(old_start),
-      old_range = tonumber(old_range) or 0,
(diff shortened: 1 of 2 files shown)


jakubbortlik:
Hi Sebastian! Thanks for opening the PR. I'm aware of hunk header parsing bug and the same fix is part of my [WIP](https://github.com/jakubbortlik/gitlab.nvim/blob/3088535eab31ab159d546bbe321ce176c6869942/lua/gitlab/hunks.lua#L30-L32). Regarding the modification type detection, I'm playing with the idea of actually checking the diff with a three line context (`--unified=3`) because that will possibly make it easier to detect "expanded" lines - a type of "non-modification" that Gitlab itself distinguishes, so maybe I'll revert the `line_was_removed` function in the future, but I'm merging this anyway now because it will prevent some of the errors commenting produces.

Thank you especially for the tests!

PR #581, opened 2026-08-03
fix: discussion window teardown and a recursion guard

seflue:
Three unrelated bugs, all found while working on #577.

### A discussion window that can't be closed

Every so often `gld` leaves a discussion window on screen. It has to go by hand, and toggling again opens a new one beside it.

`M.close` cleared `split_visible` regardless of whether anything had been closed. From then on the plugin considered the tree gone while its window was still there, and the next toggle mounted a second one beside it.

`NuiSplit:unmount()` raises an internal loading flag before `_buf_destroy()` and `_close_window()` and clears it only once both return. An error in between leaves the object loading for good, and every later `unmount()` returns early without closing anything. One trigger is `E444: Cannot close last window`, raised inside `_close_window` when the split is the last window in the session. After that nui cannot close that tree at all.

`M.close` now closes the window itself when the split does not, and keeps `split_visible` set for as long as the window is alive; the `WinClosed` handler defers to the next tick, so the buffer wipe still fires `BufWipeout`. Whether that is the path behind the issues I observed I can't say; any error in between does it. #308 shows the same nui failure from the popup side.

A second window can also appear without any of this, when something splits the tree's window and the split keeps its buffer. That one is not addressed here, though the buffer release below does clear it up on the next toggle.

### get_root_node can hang Neovim

Every reply in the tree is a node with `type = "note"` and no `is_root`, and `get_root_node` walks those up to their root. On a node without a parent, `get_parent_id()` returns nil and the walk calls `tree:get_node(nil)`. NuiTree reads a missing id as "the node under the cursor" and hands back the node we started from. The recursion sits in tail position, so the stack never grows and nothing is raised: Neovim spins until you kill it.

`get_note_node` already returns nil on a missing parent, `get_root_node` does not. Every non-root note node currently sits under a root, but nothing enforces that. Both callers already report a nil root as an error, so returning nil is enough to make the failure visible. I ran into this while writing a test, which hung instead of failing.

### Two leaked buffers per open/close cycle

`:ls` fills with stale `gitlab` buffers over a session. `create_split_and_bufs` allocates a fresh linked/unlinked pair on every open and nothing ever released the previous one. They are listed buffers, so until now you could reach a closed discussion tree again with `:bnext`, but after this you can't.

### DIFF
--- a/lua/gitlab/actions/discussions/init.lua
+++ b/lua/gitlab/actions/discussions/init.lua
@@ -155,7 +155,11 @@ M.open = function(callback, view_type)
   -- Set autocmd to clean up state when discussions split is closed manually
   vim.api.nvim_create_autocmd("WinClosed", {
     pattern = tostring(M.split.winid),
-    callback = M.close,
+    -- M.close deletes the tree buffers. Autocmds do not nest, so only outside this callback
+    -- (hence vim.schedule) does that fire BufWipeout and run the resets above.
+    callback = function()
+      vim.schedule(M.close)
+    end,
   })
 
   -- Initialize winbar
@@ -180,10 +184,35 @@ M.open = function(callback, view_type)
   end
 end
 
----Clear the discussion state and unmounts the split.
+---Clear the discussion state and unmount the split.
 M.close = function()
-  if M.split then
+  if M.split == nil then
+    return
+  end
+  -- nui nils `split.winid` and `split.bufnr` as it tears them down, so read both while they
+  -- are still set.
+  local winid = M.split.winid
+  local split_bufnr = M.split.bufnr
+  if winid ~= nil and vim.api.nvim_win_is_valid(winid) then
+    local ok, err = pcall(vim.api.nvim_win_close, winid, true)
+    if not ok and tostring(err):find("E444") then
+      -- Last window in the session, so it needs a sibling before it can be closed.
+      vim.cmd("silent! vsplit")
+      ok = pcall(vim.api.nvim_win_close, winid, true)
+    end
+    if not ok then
+      u.notify("Could not close the discussion window", vim.log.levels.WARN)
+      return
+    end
+  end
+  -- Release nui's own buffer and augroups, which nothing else frees. Guarded so a failure
+  -- in there cannot skip the state cleanup below.
+  pcall(function()
     M.split:unmount()
+  end)
+  -- Reached only when unmount did not get that far, and nothing else would free that buffer.
+  if split_bufnr ~= nil and vim.api.nvim_buf_is_valid(split_bufnr) then
+    vim.api.nvim_buf_delete(split_bufnr, { force = true })
   end
   M.split_visible = false
   M.discussion_tree = nil
--- a/lua/gitlab/actions/discussions/init.lua
+++ b/lua/gitlab/actions/discussions/init.lua
@@ -35,6 +35,17 @@ local M = {
   unlinked_discussion_tree = nil,
 }
 
+---Delete discussion buffers to prevent leaked buffers on each M.open/M.close cycle.
+---@param split_bufnr number? Passed in because `unmount` has already nil'd `M.split.bufnr`.
+local function delete_bufs(split_bufnr)
+  -- pairs, because any of these might be nil
+  for _, bufnr in pairs({ split_bufnr, M.linked_bufnr, M.unlinked_bufnr }) do
+    if vim.api.nvim_buf_is_valid(bufnr) then
+      vim.api.nvim_buf_delete(bufnr, { force = true })
+    end
+  end
+end
+
 ---Re-fetch all discussions and re-render the relevant view.
 ---TODO: simplify the function signature - "unlinked" and "all" should not be two booleans
 ---@param unlinked boolean
@@ -210,12 +221,9 @@ M.close = function()
   pcall(function()
     M.split:unmount()
   end)
-  -- Reached only when unmount did not get that far, and nothing else would free that buffer.
-  if split_bufnr ~= nil and vim.api.nvim_buf_is_valid(split_bufnr) then
-    vim.api.nvim_buf_delete(split_bufnr, { force = true })
-  end
   M.split_visible = false
   M.discussion_tree = nil
+  delete_bufs(split_bufnr)
   winbar.cleanup_timer()
 end
 
--- a/tests/spec/discussions_shared_bufs_spec.lua
+++ b/tests/spec/discussions_shared_bufs_spec.lua
@@ -1,0 +1,72 @@
+-- The linked and unlinked buffers belong to one open, not to the session, so close() owns
+-- their release.
+
+local discussions = require("gitlab.actions.discussions")
+local draft_notes = require("gitlab.actions.draft_notes")
+local winbar = require("gitlab.actions.discussions.winbar")
+local state = require("gitlab.state")
+
+-- Without this precondition the deletion asserts below would also pass if open() never
+-- created the buffers in the first place.
+local function assert_buffers_created(linked, unlinked)
+  assert.is_true(linked ~= nil and vim.api.nvim_buf_is_valid(linked), "open() created no linked buffer")
+  assert.is_true(unlinked ~= nil and vim.api.nvim_buf_is_valid(unlinked), "open() created no unlinked buffer")
+end
+
+describe("actions/discussions buffers", function()
+  local original_rebuild_view
+
+  before_each(function()
+    -- M.open tails into draft_notes.rebuild_view, which talks to the Go server these tests
+    -- have no connection to.
+    original_rebuild_view = draft_notes.rebuild_view
+    draft_notes.rebuild_view = function() end
+  end)
+
+  after_each(function()
+    draft_notes.rebuild_view = original_rebuild_view
+    discussions.split = nil
+    discussions.split_visible = false
+    discussions.discussion_tree = nil
+    discussions.linked_bufnr = nil
+    discussions.unlinked_bufnr = nil
+    winbar.cleanup_timer()
+    state.DISCUSSION_DATA = nil
+    vim.cmd("tabnew")
+    vim.cmd("silent! tabonly")
+    vim.cmd("silent! only")
+  end)
+
+  it("Deletes both buffers when the window is closed", function()
+    vim.cmd("tabnew")
+    discussions.open()
+    local linked, unlinked = discussions.linked_bufnr, discussions.unlinked_bufnr
+    assert_buffers_created(linked, unlinked)
+
+    discussions.close()
+
+    assert.is_false(vim.api.nvim_buf_is_valid(linked), ("linked buffer %d was not deleted"):format(linked))
+    assert.is_false(vim.api.nvim_buf_is_valid(unlinked), ("unlinked buffer %d was not deleted"):format(unlinked))
+  end)
+
+  it("Leaves no buffer behind over an open/close cycle", function()
+    vim.cmd("tabnew")
+    discussions.open()
+    local first_linked, first_unlinked = discussions.linked_bufnr, discussions.unlinked_bufnr
+    assert_buffers_created(first_linked, first_unlinked)
+    discussions.close()
+
+    discussions.open()
+
+    assert.are_not.equal(first_linked, discussions.linked_bufnr)
+    assert.is_false(
+      vim.api.nvim_buf_is_valid(first_linked),
+      ("linked buffer %d of the first open leaked"):format(first_linked)
+    )
+    assert.is_false(
+      vim.api.nvim_buf_is_valid(first_unlinked),
+      ("unlinked buffer %d of the first open leaked"):format(first_unlinked)
+    )
+    discussions.close()
+  end)
+end)
--- a/lua/gitlab/actions/common.lua
+++ b/lua/gitlab/actions/common.lua
@@ -159,6 +159,11 @@ M.get_root_node = function(tree, node)
   end
   if node.type == "note_body" or node.type == "note" and not node.is_root then
     local parent_id = node:get_parent_id()
+    -- `tree:get_node(nil)` falls back to the node under the cursor, which can be this very
+    -- node again. Because this is a tail call recursion, it would loop forever instead of overflowing.
+    if parent_id == nil then
+      return nil
+    end
     return M.get_root_node(tree, tree:get_node(parent_id))
   elseif node.is_root then
     return node
--- a/tests/spec/common_root_node_spec.lua
+++ b/tests/spec/common_root_node_spec.lua
@@ -1,0 +1,25 @@
+-- Without the nil-parent guard this test does not fail, it hangs, and the suite stops here.
+
+local NuiTree = require("nui.tree")
+local common = require("gitlab.actions.common")
+
+describe("actions/common.get_root_node", function()
+  it("Gives up on a top level node that is not marked as a root", function()
+    local bufnr = vim.api.nvim_create_buf(false, true)
+    local tree = NuiTree({
+      bufnr = bufnr,
+      nodes = { NuiTree.Node({ id = "a", text = "a", type = "note" }) },
+    })
+    tree:render()
+    -- The loop only forms if NuiTree can answer `get_node(nil)`, which it does from the
+    -- cursor of a window showing the buffer.
+    vim.api.nvim_win_set_buf(0, bufnr)
+
+    assert.is_nil(
+      common.get_root_node(tree, tree:get_node("-a")),
+      "get_root_node claimed a root for a note node that has no parent"
+    )
+
+    vim.api.nvim_buf_delete(bufnr, { force = true })
+  end)
+end)
--- a/tests/spec/discussions_orphan_window_spec.lua
+++ b/tests/spec/discussions_orphan_window_spec.lua
@@ -1,0 +1,98 @@
+-- close() closes the window itself instead of leaving that to NuiSplit, which gives up on
+-- the last window of a session and ignores every later unmount once one has failed. These
+-- tests check that the window and nui's own buffer are gone afterwards and that
+-- `split_visible` says so.
+
+local discussions = require("gitlab.actions.discussions")
+local draft_notes = require("gitlab.actions.draft_notes")
+local winbar = require("gitlab.actions.discussions.winbar")
+local state = require("gitlab.state")
+
+---Register a split with the given unmount behaviour, in a window of its own.
+---@param unmount fun(split: table)
+---@return integer winid
+---@return integer bufnr The scratch buffer NuiSplit allocates on mount
+local function arrange(unmount)
+  vim.cmd("tabnew")
+  vim.cmd("split")
+  local winid = vim.api.nvim_get_current_win()
+  local bufnr = vim.api.nvim_create_buf(false, true)
+  discussions.split = { winid = winid, bufnr = bufnr, unmount = unmount }
+  discussions.split_visible = true
+  return winid, bufnr
+end
+
+describe("actions/discussions.close", function()
+  after_each(function()
+    discussions.split = nil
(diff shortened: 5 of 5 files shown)


lua/gitlab/actions/discussions/init.lua
```diff
@@ -182,11 +199,27 @@ end
...
-  if M.split then
+  if M.split == nil then
+    return
+  end
+  local winid = M.split.winid
+  -- NuiSplit raises an internal loading flag before destroying buffer and window and lowers
+  -- it only on the way out, so an error in between leaves the object loading forever and
+  -- every later unmount returns early without closing anything. Close the window directly
+  -- when the split does not, and keep `split_visible` set for as long as the window is on
+  -- screen: claiming it is gone makes the next toggle open a second one beside it.
```
jakubbortlik: I suggest splitting the comment and moving individual parts to the corresponding code sections. Maybe it's just me, but I had to read the whole several times before I understood that "keep `split_visible` set for as long as the window is on screen" corresponds to the return on line 217.

lua/gitlab/actions/discussions/init.lua
```diff
@@ -155,7 +167,12 @@ M.open = function(callback, view_type)
   -- Set autocmd to clean up state when discussions split is closed manually
   vim.api.nvim_create_autocmd("WinClosed", {
     pattern = tostring(M.split.winid),
-    callback = M.close,
+    -- WinClosed fires while the window is still on its way out, so the teardown can only
+    -- run afterwards: closing it a second time from here raises, and that error is exactly
+    -- what leaves NuiSplit permanently loading.
```
jakubbortlik: I find this comment a little hard to understand:
1. Does "so the teardown can only run afterwards" refer to the need to use a `vim.schedule` call?
2. What does "closing it a second time from here" mean exactly?
3. What "that error" does it raise and how can I reproduce the NuiSplit loading permanently? I can use `:q` in the discussion tree without issues.

tests/spec/discussions_shared_bufs_spec.lua
```diff
@@ -0,0 +1,63 @@
...
+    state.DISCUSSION_DATA = nil
+    vim.cmd("tabnew")
+    vim.cmd("silent! tabonly")
+    vim.cmd("silent! only")
+  end)
+
+  it("Deletes both buffers when the window is closed", function()
+    vim.cmd("tabnew")
+    discussions.open()
+    local linked, unlinked = discussions.linked_bufnr, discussions.unlinked_bufnr
```
jakubbortlik, line 43: Do you think it would make sense to also assert that the buffers are valid before closing the discussions?

lua/gitlab/actions/discussions/init.lua
seflue: Done.

lua/gitlab/actions/discussions/init.lua
seflue: Fixed. I also went through the other comments again and made them clearer. Hope it is now easier to understand.

tests/spec/discussions_shared_bufs_spec.lua
seflue: Yes, sharpening the precondition would catch a broken `open()`. Done.

lua/gitlab/actions/discussions/init.lua
```diff
@@ -35,6 +35,17 @@ local M = {
   unlinked_discussion_tree = nil,
 }
 
+---Every open creates a fresh pair of discussion buffers, so without this each open/close
+---cycle leaks two listed buffers.
```
jakubbortlik: A nitpick: In #574 I tried to unify the docstring style to (almost) always have an imperative one-line first sentence. I suggest this wording to keep that style: 
```suggestion
---Delete discussion buffers to prevent two leaked buffers on each M.open/M.close cycle.
```

lua/gitlab/actions/discussions/init.lua
```diff
@@ -155,7 +166,11 @@ M.open = function(callback, view_type)
   -- Set autocmd to clean up state when discussions split is closed manually
   vim.api.nvim_create_autocmd("WinClosed", {
     pattern = tostring(M.split.winid),
-    callback = M.close,
+    -- Unmounting from here would close the still-closing window a second time; the
+    -- resulting error is what leaves the split stuck loading (see M.close).
```
jakubbortlik: I still have trouble understanding this comment. Is that an explanation why M.close is used here instead of using `M.split:unmount`? I'm not sure because `M.split:unmount` is eventually called within `M.close` anyway. Or is it an explanation why `vim.schedule` is used?
What is "the resulting error"? The error that we prevent here somehow? In that case the present tense in "error *is* what *leaves* the split stuck loading" is confusing. Also, I don't find `M.close` and the comments there helpful in understanding this.

Maybe the reason I have trouble understanding is that I've never had issues with closing the split or it getting "stuck loading".

lua/gitlab/actions/discussions/init.lua
```diff
@@ -182,11 +197,26 @@ end
 
 ---Clear the discussion state and unmounts the split.
```
jakubbortlik: I forgot this "s" in my #574 docs review.
```suggestion
---Clear the discussion state and unmount the split.
```

lua/gitlab/actions/discussions/init.lua
```diff
@@ -182,11 +197,26 @@ end
...
+  if M.split == nil then
+    return
+  end
+  local winid = M.split.winid
+  -- nui's unmount returns early when `_.loading` is set and clears it only at the end,
+  -- so a failure in between disables every later unmount.
+  pcall(function()
     M.split:unmount()
+  end)
+  -- The same failure can leave the window on screen.
```
jakubbortlik: I'd be a little more explicit here:
```suggestion
  -- If the window was not closed by unmount(), close it manually:
```

lua/gitlab/actions/discussions/init.lua
```diff
@@ -182,11 +197,26 @@ end
...
+  if winid ~= nil and vim.api.nvim_win_is_valid(winid) then
+    pcall(vim.api.nvim_win_close, winid, true)
+    if vim.api.nvim_win_is_valid(winid) then
+      -- Still on screen: say so, or the next toggle opens a second one beside it.
```
jakubbortlik: I'd also be more explicit here:
```suggestion
    -- If the window is still on screen, don't clear the state or the next toggle opens a second split beside it.
    if vim.api.nvim_win_is_valid(winid) then
```

jakubbortlik (commented):
I've suggested some modifications to the docstrings/comments and expressed some remaining uncertainty. I don't mean to be nitpicky, trying to understand and hopefully make the comments more useful for other contributors as well :)

lua/gitlab/actions/discussions/init.lua
seflue: I derived that from nui's `unmount`. I have since tried to close that window, but could not get `unmount` to raise in this branch's code, which is why it did not reproduce for you. The mechanism is real, but the actual failure is a different one: a buffer wiped from inside a `WinClosed` callback fires no `BufWipeout`, and the autocmds that reset `linked_bufnr` and `unlinked_bufnr` to `nil` listen for that event.

Here it's just hardening, `open` overwrites both fields anyway. In #577 it matters, because the buffers will be shared across tabs and the reset to `nil` is how the code knows they are gone.

I rewrote the comment, hope it is clearer now. It will be touched in #577 again.

lua/gitlab/actions/discussions/init.lua
seflue: Done.

lua/gitlab/actions/discussions/init.lua
seflue: Done

lua/gitlab/actions/discussions/init.lua
seflue: Done

lua/gitlab/actions/discussions/init.lua
```diff
@@ -180,13 +194,28 @@ M.open = function(callback, view_type)
...
+  pcall(function()
     M.split:unmount()
+  end)
+  -- If the window was not closed by unmount(), close it manually:
+  if winid ~= nil and vim.api.nvim_win_is_valid(winid) then
+    pcall(vim.api.nvim_win_close, winid, true)
+    -- If the window is still on screen, don't clear the state or the next toggle opens a second split beside it.
+    if vim.api.nvim_win_is_valid(winid) then
+      return
+    end
```
jakubbortlik: One more question regarding this early return: Can we ever get to the state that the window is not closed properly for some reason but `M.split` is already set to `nil` so `split_visible`, `discussion_tree`, and the buffer numbers would never be cleared by `close()`?

seflue: No. `M.split` is only ever assigned in `M.open` and never reset, so it is nil only before the first open of a session, when there is no window and nothing to clear. The state you describe, a window still on screen while `M.split` is already nil, cannot occur.

The guard is not decoration though: `reviewer/init.lua:86` calls `close()` on every reviewer start, before any tree exists, and without it the `M.split.winid` on the next line would throw. What `unmount()` clears is `split.winid` inside the object, which is why `close()` reads the window id before unmounting.

More generally, the state stays uncleared only while the window is genuinely on screen, so it never claims something untrue, and it recovers on the next `close()` as soon as the window can be closed at all (which is only possible if it is not the last window on screen). The steps below show both halves: the tree is still there, `split_visible` says so, and one `:vsplit` later `gld` clears everything.

I also have the reproduction you asked for earlier. On develop:

1. open an MR, make sure, you see a discussion tree
2. move the cursor into the tree window
3. `:only`, then `:tabonly`, so the tree is the only window in the session
4. `gld`

```
E5108: Lua: .../nui.nvim/lua/nui/split/init.lua:199: Vim:E444: Cannot close last window
stack traceback:
    [C]: in function 'nvim_win_close'
    .../nui/split/init.lua:199: in function '_close_window'
    .../nui/split/init.lua:323: in function 'unmount'
    .../gitlab/actions/discussions/init.lua:186: in function 'close'
    .../gitlab/init.lua:87: in function 'toggle_discussions'
```

`unmount()` raised between setting `_.loading` and clearing it, so from here on the split is stuck loading and every later `unmount()` returns early without closing anything. `:vsplit` and then run `gld` twice: the tree stays while the plugin flips its state anyway.

With this PR applied, the same four steps do nothing visible. Both the unmount and the manual close hit E444, the early return keeps `split_visible` true, and e.g. `:vsplit` followed by `gld` then closes the tree properly.

That also corrects my earlier guess: the `WinClosed` handler was not the trigger, the window is already invalid there and nui's `win_is_valid` guard skips the close.

I adjusted the code comments and the PR description to my latest findings.

lua/gitlab/actions/discussions/init.lua
```diff
@@ -155,7 +165,11 @@ M.open = function(callback, view_type)
   -- Set autocmd to clean up state when discussions split is closed manually
   vim.api.nvim_create_autocmd("WinClosed", {
     pattern = tostring(M.split.winid),
-    callback = M.close,
+    -- M.close wipes the discussion buffers, and a buffer wiped from inside this callback
+    -- fires no BufWipeout, so the autocmds above would never reset the bufnr fields.
```
jakubbortlik: Sorry to bother you, but I still don't understand what this comment's message is. Is it an explanation why we don't wipe the buffers with delete_bufs from this callback directly? Or an explanation why vim.schedule is used to delay the M.close execution so that it happens only after the callback exits?

lua/gitlab/actions/discussions/init.lua
```diff
@@ -180,13 +194,32 @@ M.open = function(callback, view_type)
...
+  end)
+  -- If the window was not closed successfully by unmount(), close it manually:
+  if winid ~= nil and vim.api.nvim_win_is_valid(winid) then
+    pcall(vim.api.nvim_win_close, winid, true)
+    -- Both closes fail with E444 when the tree is the last window in the session. It really
+    -- is still open then, so keep the state saying so or the next toggle opens a second
+    -- split beside it.
+    if vim.api.nvim_win_is_valid(winid) then
+      return
+    end
```
jakubbortlik: I'm having a look again and all this seems too complicated to me -- it seems like such a simple thing as closing a window is taking too much logic and too many comments, and in some cases it even doesn't do what the user is asking to do: close the window.

I suggest we get rid of the `split:unmount` (if it's unreliable anyway) and do something like [snacks.nvim](https://github.com/folke/snacks.nvim/blob/8b5f76292becf9ad76ef1507cbdcec64a49ff3f4/lua/snacks/win.lua#L561-L569):
```lua
    if winid and vim.api.nvim_win_is_valid(winid) then
      local ok, err = pcall(vim.api.nvim_win_close, winid, true)
      if not ok and (err and err:find("E444")) then
        -- last window, so create a split and close the window again
        vim.cmd("silent! vsplit")
        pcall(vim.api.nvim_win_close, winid, true)
        -- maybe verify here that this time the window was actually closed?
      elseif not ok then
        u.notify(SOME SHORT MEANINGFUL ERROR MESSAGE) 
        return
      end
    end
```

seflue: You're right to question that. I have to admit that I was primarily focussed on getting my feature to run and getting rid of the most annoying side effect of the former code (a stack trace right in the user's face). And you're right about the other part too: leaving the window open when the user asked to close it is a bad trade, no matter how rare the case is.

Looking closer, setup and teardown of the discussion window are quite asymmetric. While `mount` opens the window last, `unmount` closes it last as well, instead of reversing the order. So the only step that can fail sits at the very end, which is why the loading flag can stay set, and where all the state juggling in `close` comes from.

Dropping it entirely is a different matter, though. `open` builds a `Split` and mounts it, which gives us a scratch buffer we never use (`nui/split/init.lua:144`) and two augroups (`:219-220`). `unmount` is the only thing that frees them, so dropping it leaks one buffer and two augroups per opened discussion window, for the whole session.

So we close the window your way now, and call `unmount` afterwards, when its own close step is already a no-op and cannot throw. I pushed that, including the E444 fallback and a test for the last-window case. The early return you objected to is gone, the comment about the loading flag went with it, and `close` came out shorter overall.

What I would like to keep for a follow-up is the bigger question, whether we need NuiSplit here at all. I haven't looked into that deeply enough to say. And doing it now risks doing it twice: I would rather work against #577 once it is merged, so the code already covers the requirements needed for it.

After that I would be happy to clean up setup and teardown properly. And much more motivated. :slightly_smiling_face:

lua/gitlab/actions/discussions/init.lua
seflue: The second one: it explains the `vim.schedule`. And your two readings are the same decision from two ends, because [`M.close` is what deletes the buffers](https://github.com/harrisoncramer/gitlab.nvim/blob/06bb8d3b9fb056be0ac4d1cc9baab5779b611221/lua/gitlab/actions/discussions/init.lua#L39-L46).

The mechanism I failed to name is `:h autocmd-nested`: autocommands do not nest, so a buffer wiped from inside this callback fires `BufWipeout` into the void and [the two autocmds above](https://github.com/harrisoncramer/gitlab.nvim/blob/06bb8d3b9fb056be0ac4d1cc9baab5779b611221/lua/gitlab/actions/discussions/init.lua#L152-L163) never reset `linked_bufnr` and `unlinked_bufnr`. Deferring `M.close` moves the wipe out of the autocmd, where the event fires normally.

I also have to correct myself: I told you earlier this was just hardening because `open` overwrites both fields anyway. It is not. `nil` is how the plugin knows the buffers are gone, and two independent paths read it while the tree is closed.

One of them is the reviewer, and it is the one a user is far more likely to trigger. [`refresh_diagnostics`](https://github.com/harrisoncramer/gitlab.nvim/blob/06bb8d3b9fb056be0ac4d1cc9baab5779b611221/lua/gitlab/actions/discussions/init.lua#L100-L102) is bound to `DiffviewViewEnter`, so it runs every time you enter the diff tab. It calls `add_empty_titles`, which [writes a placeholder line](https://github.com/harrisoncramer/gitlab.nvim/blob/06bb8d3b9fb056be0ac4d1cc9baab5779b611221/lua/gitlab/actions/common.lua#L82-L91) such as "No Notes (Unlinked Discussions) for this MR" into whichever of the two tree buffers currently holds no comments. Most MRs have no unlinked notes, so that write happens nearly every time. And if the buffer number is stale, it writes into a buffer that has been wiped: `nvim_buf_set_lines` raises `Invalid buffer id` with a traceback, once per entry into the diff tab, until you toggle the tree open again. That would explain the `Invalid buffer id` tracebacks I have been seeing.

The other is [`refresh_data`](https://github.com/harrisoncramer/gitlab.nvim/blob/06bb8d3b9fb056be0ac4d1cc9baab5779b611221/lua/gitlab/init.lua#L98-L101), which ends in the two tree rebuilds ([:507](https://github.com/harrisoncramer/gitlab.nvim/blob/06bb8d3b9fb056be0ac4d1cc9baab5779b611221/lua/gitlab/actions/discussions/init.lua#L507), [:546](https://github.com/harrisoncramer/gitlab.nvim/blob/06bb8d3b9fb056be0ac4d1cc9baab5779b611221/lua/gitlab/actions/discussions/init.lua#L546)) and throws [unconditionally](https://github.com/harrisoncramer/gitlab.nvim/blob/06bb8d3b9fb056be0ac4d1cc9baab5779b611221/lua/gitlab/actions/discussions/init.lua#L517). I concluded both from reading the code and have not taken the time to reproduce them yet.

lua/gitlab/actions/discussions/init.lua
jakubbortlik: Thanks for your patience with the lengthy review. Doing it in my spare time...

"whether we need NuiSplit here at all." That's what I was asking myself too. Let's do that later as you say.

jakubbortlik:
~One last thing (promise 🤞) . Last time I used tour branch the gld gld cycle of closing and reopening the split still seemed to leave some orphaned buffers behind (not sure though if that was gitlab.nvim's fault or some other plugin, I didn't try with a clean install and only the necessary dependencies).~

I cannot reproduce this anymore. Closing.

tests/spec/discussions_shared_bufs_spec.lua
```diff
@@ -0,0 +1,72 @@
...
+    assert_buffers_created(linked, unlinked)
+
+    discussions.close()
+
+    assert.is_false(vim.api.nvim_buf_is_valid(linked), ("linked buffer %d was not deleted"):format(linked))
```
jakubbortlik, line 48: The Nui.split creates its own buffer (accessible through [split.bufnr](https://github.com/seflue/gitlab.nvim/blob/20852e3cf01b4c4924d0b3b354d7fafa3fba464f/lua/gitlab/actions/discussions/init.lua#L136)). I guess we should check that this buffer is also removed, but maybe this check belongs into the other spec file related to the orphaned window: `tests/spec/discussions_orphan_window_spec.lua`, as we should make sure the buffer is removed even when the `split:nmount` fails.

seflue: Done. `close()` frees the buffer itself now, because `mount` has the same flaw we worked around in `unmount`: if `_open_window()` raises, `_.loading` stays set and every later `unmount()` returns early.

lua/gitlab/actions/discussions/init.lua
```diff
@@ -216,6 +226,7 @@ M.close = function()
   end
```
jakubbortlik: I think you could move this piece into the `delete_bufs` helper function.

Maybe you could also consider replacing the three structurally identical if blocks with a loop over the buffer numbers.

seflue: Done.

jakubbortlik:
Hi Sebastian, thanks for the fixes and your patience with the review.

Issue #478, opened 2025-02-19
Ranged comments don't set range correctly


jakubbortlik:
## Bug Description

The line range of a comment is not properly applied in Gitlab. If I create a ranged comment in Gitlab online, it actually contains the range information (it's the second comment in the screenshot below).

The missing range information is not just cosmetic it also makes it impossible to properly insert a suggestion in the `gitlab.nvim` "ranged" comment. If I want to modify it and click "Insert suggestion" it pastes this:

````
```suggestion:-0+0
abc
```
````
Instead of the correct:
````
```suggestion:-1+0
defghi
abc
```
````

The problem is that setting this correctly from within `gitlab.nvim` may currently not be possible due to limitations of the [Gitlab API](https://docs.gitlab.com/api/discussions/#parameters-for-multiline-comments) and of the [go library](https://gitlab.com/gitlab-org/api/client-go/-/blob/89e6601b09ec7702973fda7d24358e7bc59e197c/discussions.go#L756-760) that only seem to allow setting the line code and line position type ("old" vs "new").

## Screenshots
First image shows a comment created in `gitlab.nvim`, the second one is directly from `Gitlab`:
![Image](https://github.com/user-attachments/assets/1f46ae53-2c85-4a6a-b136-7f727f10c4a6)

jakubbortlik:
I've opened a [MR in client-go](https://gitlab.com/gitlab-org/api/client-go/-/merge_requests/2254). I've tested in a local experimental branch of `gitlab.nvim` that this modification to `client-go` enables setting notes like "Comment on lines 1 to 2" on discussions.

However, the way how the [LinePositionOptions](https://github.com/harrisoncramer/gitlab.nvim/blob/65179abebdb410821909588bf5aee051061b4eeb/cmd/app/comment_helpers.go#L77-L84) are calculated will have to be fixed, because currently, the plugin creates incorrect line ranges like this (`-1 to -2` instead of `1 to 2`):
![Image](https://github.com/user-attachments/assets/d08e1dbc-d4bb-4832-9eb3-b61c7683013f)

One part of the issue is that for unchanged lines of code, the [Type](https://github.com/harrisoncramer/gitlab.nvim/blob/65179abebdb410821909588bf5aee051061b4eeb/cmd/app/comment_helpers.go#L78) has to be the empty string, rather than `"old"`. I suspect, the incorrect way how the `LinePositionOptions` are currently prepared may also lead to errors like in #386.

[...]

PR #582, opened 2026-08-07
fix: get correct location data for comments efficiently

### LINKED ISSUE #386: Trying to make a comment raises the error: line_code=>["can't be blank", "must be a valid line code"]

jakubbortlik:
This PR fixes #386 and #478.

Apart from sending the correct data to Gitlab (correct line codes, line numbers and modification types), this plugin now creates these data more efficiently, with a single git diff call and with a single iteration through the diff. The PR removes many (often nearly identical) helper functions and adds tests for the newly added or modified functions.

### DIFF
--- a/cmd/app/comment_helpers.go
+++ b/cmd/app/comment_helpers.go
@@ -7,8 +7,18 @@ import (
 	gitlab "gitlab.com/gitlab-org/api/client-go"
 )
 
-/* LinePosition represents a position in a line range. Unlike the Gitlab struct, this does not contain LineCode with a sha1 of the filename */
-type LinePosition struct {
+/* PositionInfo represents one endpoint (start or end) of a line range, as sent by the Lua
+* plugin. Unlike the Gitlab struct, it has no LineCode - Lua can't compute a sha1, so
+* buildCommentPosition computes one below from OldLine and NewLine.
+*
+* OldLine and NewLine are always real, non-nil integers, even when Type is "old" or "new"
+* and only one side actually has a line. On the side that doesn't, the value is a position
+* marker, not a claim that a line exists there: it's wherever that side's cursor was
+* sitting when the other side's line was found. LineCode is always built from this
+* unzeroed pair; buildCommentPosition separately zeroes the inapplicable side before
+* setting it on the request's LineRange.{Start,End}.{OldLine,NewLine} - see
+* zeroInapplicableLine. */
+type PositionInfo struct {
 	Type    string `json:"type"`
 	OldLine int64  `json:"old_line"`
 	NewLine int64  `json:"new_line"`
@@ -16,8 +26,8 @@ type LinePosition struct {
 
 /* LineRange represents the range of a note. */
 type LineRange struct {
-	StartRange *LinePosition `json:"start"`
-	EndRange   *LinePosition `json:"end"`
+	Start *PositionInfo `json:"start" validate:"required"`
+	End   *PositionInfo `json:"end" validate:"required"`
 }
 
 /* PositionData represents the position of a comment or note (relative to a file diff) */
@@ -30,7 +40,7 @@ type PositionData struct {
 	BaseCommitSHA  string     `json:"base_commit_sha"`
 	StartCommitSHA string     `json:"start_commit_sha"`
 	Type           string     `json:"type"`
-	LineRange      *LineRange `json:"line_range,omitempty"`
+	LineRange      *LineRange `json:"line_range" validate:"required_with=FileName"`
 }
 
 /* RequestWithPosition is an interface that abstracts the handling of position data for a comment or a draft comment */
@@ -42,48 +52,65 @@ type RequestWithPosition interface {
 func buildCommentPosition(commentWithPositionData RequestWithPosition) *gitlab.PositionOptions {
 	positionData := commentWithPositionData.GetPositionData()
 
-	// If the file has been renamed, then this is a relevant part of the payload
-	oldFileName := positionData.OldFileName
-	if oldFileName == "" {
-		oldFileName = positionData.FileName
-	}
-
 	opt := &gitlab.PositionOptions{
 		PositionType: &positionData.Type,
 		StartSHA:     &positionData.StartCommitSHA,
 		HeadSHA:      &positionData.HeadCommitSHA,
 		BaseSHA:      &positionData.BaseCommitSHA,
 		NewPath:      &positionData.FileName,
-		OldPath:      &oldFileName,
+		OldPath:      &positionData.OldFileName,
 		NewLine:      positionData.NewLine,
 		OldLine:      positionData.OldLine,
 	}
 
-	if positionData.LineRange != nil {
-		shaFormat := "%x_%d_%d"
-		startFilenameSha := fmt.Sprintf(
-			shaFormat,
-			sha1.Sum([]byte(positionData.FileName)),
-			positionData.LineRange.StartRange.OldLine,
-			positionData.LineRange.StartRange.NewLine,
-		)
-		endFilenameSha := fmt.Sprintf(
-			shaFormat,
-			sha1.Sum([]byte(positionData.FileName)),
-			positionData.LineRange.EndRange.OldLine,
-			positionData.LineRange.EndRange.NewLine,
-		)
-		opt.LineRange = &gitlab.LineRangeOptions{
-			Start: &gitlab.LinePositionOptions{
-				Type:     &positionData.LineRange.StartRange.Type,
-				LineCode: &startFilenameSha,
-			},
-			End: &gitlab.LinePositionOptions{
-				Type:     &positionData.LineRange.EndRange.Type,
-				LineCode: &endFilenameSha,
-			},
-		}
+	shaFormat := "%x_%d_%d"
+	startFilenameSha := fmt.Sprintf(
+		shaFormat,
+		sha1.Sum([]byte(positionData.FileName)),
+		positionData.LineRange.Start.OldLine,
+		positionData.LineRange.Start.NewLine,
+	)
+	endFilenameSha := fmt.Sprintf(
+		shaFormat,
+		sha1.Sum([]byte(positionData.FileName)),
+		positionData.LineRange.End.OldLine,
+		positionData.LineRange.End.NewLine,
+	)
+
+	startOldLine, startNewLine := zeroInapplicableLine(positionData.LineRange.Start)
+	endOldLine, endNewLine := zeroInapplicableLine(positionData.LineRange.End)
+
+	opt.LineRange = &gitlab.LineRangeOptions{
+		Start: &gitlab.LinePositionOptions{
+			Type:     &positionData.LineRange.Start.Type,
+			LineCode: &startFilenameSha,
+			OldLine:  &startOldLine,
+			NewLine:  &startNewLine,
+		},
+		End: &gitlab.LinePositionOptions{
+			Type:     &positionData.LineRange.End.Type,
+			LineCode: &endFilenameSha,
+			OldLine:  &endOldLine,
+			NewLine:  &endNewLine,
+		},
 	}
 
 	return opt
+}
+
+/* zeroInapplicableLine returns a line_range endpoint's OldLine/NewLine with the side
+* that its Type doesn't apply to zeroed out: NewLine for a deleted ("old") line, OldLine
+* for an added ("new") line. Both stay real for an unmodified ("") or "expanded" line.
+* The unzeroed pair is still what the LineCode hash above is computed from - Gitlab
+* expects LineCode to encode the real old/new correspondence even when the displayed
+* OldLine or NewLine is zeroed. */
+func zeroInapplicableLine(position *PositionInfo) (oldLine int64, newLine int64) {
+	oldLine, newLine = position.OldLine, position.NewLine
+	switch position.Type {
+	case "old":
+		newLine = 0
+	case "new":
+		oldLine = 0
+	}
+	return oldLine, newLine
 }
--- a/lua/gitlab/actions/common.lua
+++ b/lua/gitlab/actions/common.lua
@@ -267,6 +267,7 @@ M.get_line_numbers_for_range = function(old_line, new_line, start_line_code, end
   elseif new_line ~= nil then
     local range = new_end_line - new_start_line
     -- Force start_line to be greater than 0
+    -- TODO: use `math.max(new_line - range, 1)` instead
     local start_line = (new_line - range > 0) and (new_line - range) or 1
     return start_line, new_line, true
   else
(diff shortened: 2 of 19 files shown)


jakubbortlik:
> This PR fixes #386 and #478.
> 
> Apart from sending the correct data to Gitlab (correct line codes, line numbers and modification types), this plugin now creates these data more efficiently, with a single git diff call and with a single iteration through the diff. The PR removes many (often nearly identical) helper functions and adds tests for the newly added or modified functions.

Follow-ups will include:
- review of how diagnostics for comments are placed
- review of jumping to reviewer and jumping to file
- refactor of the various functions for creating comments to just one parametrized function (already in #504)
- a suggestions keymap in the comments popup
- further work on #504

cmd/app/comment_helpers.go
```diff
@@ -7,17 +7,27 @@ import (
...
 	OldLine int64  `json:"old_line"`
 	NewLine int64  `json:"new_line"`
 }
 
 /* LineRange represents the range of a note. */
 type LineRange struct {
-	StartRange *LinePosition `json:"start"`
-	EndRange   *LinePosition `json:"end"`
+	Start *PositionInfo `json:"start"`
+	End   *PositionInfo `json:"end"`
```
seflue:
```suggestion
	Start *PositionInfo `json:"start" validate:"required"`
	End   *PositionInfo `json:"end" validate:"required"`
```
Would it make sense to validate the pointer, so a bad payload from the lua client gets a 400 instead of nil-panicking in `buildCommentPosition`?

cmd/app/comment_helpers.go
```diff
@@ -30,7 +40,7 @@ type PositionData struct {
 	BaseCommitSHA  string     `json:"base_commit_sha"`
 	StartCommitSHA string     `json:"start_commit_sha"`
 	Type           string     `json:"type"`
-	LineRange      *LineRange `json:"line_range,omitempty"`
+	LineRange      *LineRange `json:"line_range"`
```
seflue:
```suggestion
	LineRange      *LineRange `json:"line_range" validate:"required_with=FileName"`
```
Also protect `LineRange` against a wrong payload from the lua client.

lua/gitlab/actions/common.lua
```diff
@@ -267,6 +267,7 @@ M.get_line_numbers_for_range = function(old_line, new_line, start_line_code, end
   elseif new_line ~= nil then
     local range = new_end_line - new_start_line
     -- Force start_line to be greater than 0
+    -- TODO: use `math.max(new_line - range, 1)` instead
```
seflue, line 270: Why the TODO comment instead of fixing it right away?

lua/gitlab/annotations.lua
```diff
@@ -98,21 +107,12 @@
 ---@field lnum number
 ---@field buffer number?
 
----@class LineRange
+---@class ReviewerData
+---@field file_name string
+---@field old_file_name string Relevant for renamed files only, the name of the file in the previous commit
```
seflue: Could it be that some annotations are missing here?

`get_reviewer_data` returns seven keys: `old_sha` and `new_sha` are missing here.

The comment for `old_file_name` seems a bit off now that you always set it at [reviewer/init.lua:208](https://github.com/harrisoncramer/gitlab.nvim/pull/582/changes#diff-c00a20ab47f9b97ed6b8272cae8d639201cbf8e7f51804939062ec9566984963R208)

lua/gitlab/hunks.lua
```diff
@@ -36,249 +30,88 @@ M.parse_possible_hunk_headers = function(line)
...
-
----Parse the lines from a diff and return the index of the next hunk, when provided an
----initial index.
----@param lines string[]
----@param i integer
----@return integer?
-local next_hunk_index = function(lines, i)
-  for j, line in ipairs(lines) do
+  local current_hunk = nil
+  for line in diff:gmatch("[^\r\n]+") do
```
seflue, line 50: If my analysis is correct the following could happen:

Commenting on a range ending on a blank line at the last hunk of the files diff could cause the comment to get attached at a different line then selected. The offset would be however many lines that hunks adds or removes.

A blank line consists only of a space prefix. `vim.fn.trim` in `run_system` at [git.lua:11](https://github.com/harrisoncramer/gitlab.nvim/blob/30bf262fc9e74a5340c70df578d5d6ee813974d9/lua/gitlab/git.lua#L11) strips trailing whitespace from the whole output, so that line never becomes part of `hunk.lines`. Because the `@@` header still declares the full range, `get_line_position` enters the hunk and walk it's body, never matches and hits the `"expanded"` return at the bottom. That return derives `old_line` from `net_change_before`, which is never incremented for the hunk you are inside, so it is off by that hunk's delta.

`vim.fn.trim(vim.fn.system(command), "\r\n")` in `run_system` fixes it, so the line survives, the walk matches it as context, and the type comes back `""` with the right numbers.

Another subtle cause could be if someone sets `diff.suppressBlankEmpty=true` in their git config. Passing `-c diff.suppressBlankEmpty=false` in the git command would rule it out.

lua/gitlab/reviewer/init.lua
```diff
@@ -34,6 +33,15 @@ M.open = function()
   require("gitlab.git_async").check_current_branch_up_to_date_on_remote()
   local git = require("gitlab.git")
 
+  -- The rename threshold used by Gitlab (through Gitaly) is 30%, see
+  -- https://gitlab.com/gitlab-org/gitaly/-/blob/db39e26f8f8a8da62e2c2db00325cf51315c89db/internal/gitaly/service/diff/commit_diff.go#L64-64
+  -- https://gitlab.com/gitlab-org/gitaly/-/blob/0e81e24ae1f650c242670eb7bf66c4b4b91b7813/internal/gitaly/service/diff/find_changed_paths.go#L116-116
+  -- dlyongemallo/diffview.nvim can be configured to respect this:
+  M._original_diffview_config = require("diffview.config").get_config()
```
seflue: Minor but could surprise users of Diffview: you capture the original config unconditionally and when restoring you keep the captured value in `M._original_diffview_config`. If the user closes the Review diffview by hand, close is not run, so the config override leaks into the users setting. And because you capture it, a reopen would capture gitlab.nvims override as the users setting and that will be kept for the rest of the session.

If you check it for `nil` before capturing it and reset to `nil` after restore in line 111 it would at least become self-healing when reopen and closing the gitlab.nvim's diffview properly.

seflue (changes requested):
Thanks, Jakub. Calculating the diff the same way GitLab does makes a lot of sense, and it takes a whole class of edge cases out of the code.

Most of what I found is minor and noted inline. The one you should consider the most is the blank-line case in hunks.lua: a comment could end up attached to a different line than the one selected.

lua/gitlab/actions/common.lua
jakubbortlik: I wanted to keep this PR focused on creating comments. I noticed this piece of code and just made a note for a future PR(s) that will deal with jumping to reviewer and placing diagnostics.

lua/gitlab/reviewer/init.lua
jakubbortlik: Yes, this is a half-baked hack and it doesn't even work :grin:. For some reason even with this diffview setup in place just before the `DiffviewOpen` command is run, the value when running [diff_file_list](https://github.com/dlyongemallo/diffview-plus.nvim/blob/6d7149be18917fe89af3b91d1dcd28260805084b/lua/diffview/vcs/utils.lua#L154) was still set to value in my gitlab.nvim config. So I removed this and replaced with a recommendation for config.

I've also opened an issue in the [diffview-plus.nvim](https://github.com/dlyongemallo/diffview-plus.nvim/issues/301) repo to allow specifying the rename threshold for each command.

cmd/app/comment_helpers.go
jakubbortlik: Thanks for pointing this out. I've added the validation.

cmd/app/comment_helpers.go
jakubbortlik: Thanks. Validation added. This is now also covered with [tests](https://github.com/jakubbortlik/gitlab.nvim/blob/cabf86b97bebafd30c8724f283ed36e7b8745a40/cmd/app/middleware_test.go#L114-L148).

lua/gitlab/annotations.lua
jakubbortlik: Yes, I forgot to add these two. Fixed.

lua/gitlab/hunks.lua
jakubbortlik: Thanks for finding this and suggesting the solutions. I've applied them both. Actually, in my other PR, I'm using something a [little different](https://github.com/jakubbortlik/gitlab.nvim/blob/ed93eac24e16202a84840981cddcf49838ead7c2/lua/gitlab/git.lua#L12) because (if I remember correctly) there I want to keep empty lines at ends of files when getting the file contents in a given revision. But I'll keep your suggested solution for now.

cmd/app/comment_helpers.go
```diff
@@ -30,7 +40,7 @@ type PositionData struct {
 	BaseCommitSHA  string     `json:"base_commit_sha"`
 	StartCommitSHA string     `json:"start_commit_sha"`
 	Type           string     `json:"type"`
-	LineRange      *LineRange `json:"line_range,omitempty"`
+	LineRange      *LineRange `json:"line_range" validate:"required_with=FileName"`
```
seflue, line 43: Thanks for adding this.

What I didn't think of in the initial review: GitLab's position object has a third `position_type`, `file`, which carries paths and no line numbers at all. In their generated OpenAPI spec the only required members are:

- `base_sha`
- `start_sha`
- `head_sha`
- `position_type`

Optional:
- `new_line`,
- `old_line`
- `line_range`

That is what a file-level comment looks like on the wire. Since the tag makes `file_name` imply `line_range`, we would now reject such a payload. Nothing sends it today - `lua/gitlab/actions/comment.lua:102` hardcodes `type = "text"` - but we should take it into consideration when we add file-level comments.

Minor, while I was in there: it seems no test currently covers `file_name` set with `line_range` missing.

seflue (approved):
I just left one inline-comment to consider, non-blocking.

lua/gitlab/reviewer/init.lua
jakubbortlik: David, the maintainer of `diffview-plus.nvim`, has self-assigned the [issue](https://github.com/dlyongemallo/diffview-plus.nvim/issues/301). In my experience he's very efficient with implementing features, so I'll keep this PR open for a bit and maybe we'll be able to merge it with the functionality in place which would be preferable to suggesting some user configuration and replacing it with different instructions soon afterwards.

jakubbortlik: The `--rename-threshold` option for `DiffviewOpen` and `DiffviewFileHistory` has been merged to main of `diffview-plus.nvim`. The option is silently ignored by sindrets' version of the plugin (and by older versions of `diffview-plus.nvim`), so we don't need to handle it specially. I've just modified the documentation.

cmd/app/comment_helpers.go
jakubbortlik: I've added a test: https://github.com/jakubbortlik/gitlab.nvim/blob/af29555450c703ae80767e2b1238fb68f526917c/cmd/app/middleware_test.go#L134-L153. The way the PositionData is validated will have to change when `type="file"` is implemented. But I don't find it necessary now - there are workarounds (like commenting on the first line of the file with a suitable comment), and there are more urgent issues.

Anyway, thanks for the thorough review!

cmd/app/comment_helpers.go
```diff
@@ -42,48 +52,65 @@ type RequestWithPosition interface {
 func buildCommentPosition(commentWithPositionData RequestWithPosition) *gitlab.PositionOptions {
 	positionData := commentWithPositionData.GetPositionData()
 
-	// If the file has been renamed, then this is a relevant part of the payload
-	oldFileName := positionData.OldFileName
-	if oldFileName == "" {
-		oldFileName = positionData.FileName
-	}
```
jakubbortlik, line 49: This is not needed as the lua plugin now always fills `OldFileName` for renamed and non-renamed files alike.

PR #584, opened 2026-08-19
Draft: implement indentation as virtual text

jakubbortlik:
This PR re-implements the discussion tree indentation as virtual text. This has several benefits:
- selections copied from the discussion tree don't contain the indentation characters
- markdown syntax highlighting works for code blocks that mark the language, e.g., \`\`\`python
- we get rid of most of the syntax regexes in `after/syntax/gitlab.vim`.

It already works well, but needs a thorough review and should add unit tests.


lua/gitlab/actions/discussions/init.lua
```diff
@@ -35,6 +35,22 @@ local M = {
...
+---@param bufnr number
+local function attach_markdown_treesitter(bufnr)
+  vim.treesitter.language.register("markdown", "gitlab")
+  local ok = pcall(vim.treesitter.start, bufnr, "markdown")
+  if ok then
+    -- vim.treesitter.start() clears 'syntax' on the buffer to avoid double-highlighting, but
+    -- that also wipes the custom GitlabUsername/GitlabDate/etc. groups from
+    -- after/syntax/gitlab.vim, which aren't defined by the markdown parser. Reload them on top.
+    vim.bo[bufnr].syntax = "gitlab"
+  end
```
jakubbortlik, line 55: Won't this also reload the markdown syntax? Most of the groups are gone anyway...

PR #587, opened 2026-09-10
fix: jump to reviewer from comment pane

Sackbuoy:
# Overview
when calculating the line range in the lua function `get_line_numbers_for_range` in `./lua/gitlab/actions/common.lua`, there is no nil guard for new_end_line or new_start_line, so a single line comment seems to cause it to fail. This change is intended to guard against nil for those values and allow the `range` of the comment to be 0 if one of those values is found to be nil

## Expected Behavior:
When performing a code review with the comments/discussions pane focused, hitting the `Jump to comment location in file` keybind(`o` by default) should focus the file and line pertaining to the comment in the diff hunk pane. 

## Current Behavior:
Confirmed behavior on both v4.1.1, and the latest `develop` branch - Jumping to comment location in file produces the following error:
```
E5108: Lua: ...g/nvim/plugged/gitlab.nvim/lua/gitlab/actions/common.lua:268: attempt to perform arithmetic on local 'new_end_line' (a nil value)
stack traceback:
        ...g/nvim/plugged/gitlab.nvim/lua/gitlab/actions/common.lua:268: in function 'get_line_numbers_for_range'
        ...g/nvim/plugged/gitlab.nvim/lua/gitlab/actions/common.lua:285: in function 'get_line_number_from_node'
        ...g/nvim/plugged/gitlab.nvim/lua/gitlab/actions/common.lua:306: in function 'jump_to_reviewer'
        ...gged/gitlab.nvim/lua/gitlab/actions/discussions/init.lua:654: in function <...gged/gitlab.nvim/lua/gitlab/actions/discussions/init.lua:652>
```

Build/Environment info:
```
gitlab.nvim v4.1.1
zsh 5.9.2 (aarch64-apple-darwin25.6.0)
NVIM v0.12.1
Build type: Release
LuaJIT 2.1.1741730670
Run "nvim -V1 -v" for more info
```

Lua and Go tests still pass, linters came back clean - fix confirmed working when applied both to the `develop` branch and to `v4.1.1`

This is my first contribution, so let me know if anything else is needed - happy to discuss whatever questions come up


jakubbortlik:
Hi Cameron, thanks for the contribution. Before merging, I'd like to debug this a little more. You say
> so a single line comment seems to cause it to fail
This should not be the case on `develop` - every comment on develop is now a ranged comment, even one-line comments. The `nil` value must come from an unsuccessful [match on the line code](https://github.com/harrisoncramer/gitlab.nvim/blob/80edb2767a6c4fb60699996d5c4c456a7f75b17f/lua/gitlab/indicators/common.lua#L60-L61). Could you please share the line code of the comment for which jump_to_reviewer fails?

You can press `<leader>p` on the comment and this will print the tree node where you can find the `range` like this:
```lua
  range = {
    ["end"] = {
      line_code = "626a747c8694eaae66f2afb3d1a745b045782c70_0_8",
      new_line = 8,
      old_line = 0,
      type = "new"
    },
    start = {
      line_code = "626a747c8694eaae66f2afb3d1a745b045782c70_0_8",
      new_line = 8,
      old_line = 0,
      type = "new"
    }
  },
```
Maybe the actual bug lies somewhere else - the line code may not be correctly created.

